Issue #42, opened 2014-03-14
Add a useful mechanism to use loglevel in sites that are already using console.log, and want to add compatibility


pimterry:
`log.replaceConsole()`, for example, to make console.X then work reliably, even if you're in IE etc.

Doogiemuc:
I stumbled upon the same quest.  While it is possible to replace/redirect console.log (and friends) to logLevel, the remaining problem is that the browser then shows the logoutputs as if they'd be coming from the loglevel module and not from where they originally were called.

(And no bind(), call() or apply() do not help. I tried everything :-(

PR #49, opened 2014-08-03
added ability to replace console

VizuaaLOG:
This push implements two functions allowing loglevel to replace window.console but also restore it if required. I have done a manual test and both functions worked correctly.

This is to add the feature requested here: #42

### DIFF
--- a/lib/loglevel.js
+++ b/lib/loglevel.js
@@ -203,6 +203,16 @@
             return self;
         };
 
+		// Replace console.log with loglevel https://github.com/pimterry/loglevel/issues/42
+		var _console = window.console;
+		self.replaceConsole = function() {
+			window.console = self;
+		}
+
+		self.restoreConsole = function() {
+			window.console = _console;
+		}
+
         loadPersistedLevel();
         return self;
     }));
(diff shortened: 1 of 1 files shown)


r4j4h:
It's been a while on this.. does anyone still feel this would still be useful? Let me know and I will put together a pull request if so.

pimterry:
Yes, definitely useful, if you can find a way make it work. Feel free to have a go! We will need to have some tests in place too though to finish it, and we need to make sure the solution works nicely and reliably, as in my previous comment.

Don't want that to be too intimidating though. If you want to open a PR with any progress at all that'd be great, and we can go from there :-)

r4j4h:
Just reporting in that I have not had time to work on this but did a little bit of research I can share:  Potential danger with changing console by calls to [this binding function](https://github.com/pimterry/loglevel/blob/198423a28bd1ad4159ac0f0f9f0cfe30b7a07aa7/lib/loglevel.js#L32). I believe we can test by mocking two separate console objects and logging before and after switching between them, verifying that only the correct one is called and the incorrect one is not called each time. If that sounds satisfactory to everyone then I will hopefully get to it in the next few days. :)

pimterry:
Been quiet for a long while now, so closing this. Feel free to bring this up to date and reopen it if you're still interested.

PR #61, opened 2015-01-15
Updated CDN reference to the latest version

artemyarulin:
It was pointing to 0.0.6


pimterry:
Good catch, thanks for sorting that!

PR #62, opened 2015-01-15
setLevel call added to plugin example

artemyarulin:
Plugin example extended with additional required setLevel call in order to apply it


artemyarulin:
I guess it should be enough, what do you think?

pimterry:
Sorry for the delay. Yep looks good, merged, thanks!

PR #63, opened 2015-01-16
Update outdated dependencies

mroderick:
I've tried to create sensible commits, so it's probably easier to read this PR one commit at a time.
#### Switch spies/stubs to Sinon.JS

Jasmine's spies are not very elegant, and they've not become better in Jasmine 2.
The log spamming the console during test runs is a testament to that. It might be that I've missed some minor detail, or that it [just wasn't documented that well](http://jasmine.github.io/2.0/upgrading.html)

I would suggest that we switch to using Sinon.JS instead. It's battle hardened and is well documented.

I've not created any commits for this yet, as I don't want to waste effort if it's not a welcome change.
#### Intermittent test failure

On my machines, there's something weird going on with `testHelpers.js` and Jasmine's spies.

```
>> Error: require: 'undefined' is not an object (evaluating 'realLogMethod.calls.count') at
>> _SpecRunner.html:21
>> .grunt/grunt-contrib-jasmine/require.js:12 v
>> .grunt/grunt-contrib-jasmine/require.js:19
>> .grunt/grunt-contrib-jasmine/require.js:23
>> .grunt/grunt-contrib-jasmine/require.js:8
>> .grunt/grunt-contrib-jasmine/require.js:23
>> .grunt/grunt-contrib-jasmine/require.js:7 y
>> .grunt/grunt-contrib-jasmine/require.js:23
>> .grunt/grunt-contrib-jasmine/require.js:19
>> .grunt/grunt-contrib-jasmine/require.js:23
>> .grunt/grunt-contrib-jasmine/require.js:17
>> .grunt/grunt-contrib-jasmine/require.js:14 D
>> .grunt/grunt-contrib-jasmine/require.js:28
>> .grunt/grunt-contrib-jasmine/require.js:29
```

This happens on about 50% of all test runs.

This could either be from the updated `grunt-template-jasmine-requirejs` or just a bug in Jasmine's spies.

I would suggest that we add a method/property to `loglevel` itself, that we can query for the currently set level, that would obviate the `toBeAtLevel` test helper entirely.


mroderick:
@pimterry do you have any comments for this pull request?

Would you be ok with switching to `Sinon.JS` for stubs and spies?
Would you be ok with a method/property on `loglevel` that indicates the current status set?

pimterry:
Sorry, yes, I will be taking a look at this shortly. That does all sound fine in principle, I just need to find a chunk of time for a proper review here.

mroderick:
No worries. I'll add more commits to use `Sinon.JS` and expose a way of querying `logevel` for it's current level.

test/test-qunit.js
```diff
@@ -19,7 +19,7 @@ test('basic test', function() {
 
     // Check that noConflict restored the original log
     ok(typeof log === "function", "log is a function");
-    ok(log === QUnit.log, "log is Qunit.log");
+    ok(log === window.customLogFunction, "log is customLogFunction");
```
pimterry: The original reason this test was added was as a specific repro for an issue with QUnit's custom log function. That made sense to deal with the issue properly, but it's been fixed upstream in QUnit for a while now (which is why your upgrade will have highlighted this), and I think now moving this to a more generic version is indeed sensible.

If we're doing that though, this test is now totally unrelated to QUnit, so we could actually just turn this into a Jasmine test that does the same thing, within the rest of the suite, and drop the QUnit dep entirely.

test/test-helpers.js
```diff
@@ -16,18 +16,25 @@ define(function () {
...
+                var log = actual;
+                var expectedWorkingCalls = log.levels.SILENT - log.levels[expected.toUpperCase()];
+                var realLogMethod = window.console.log;
+
+                for (var ii = 0; ii < logMethods.length; ii++) {
+                    var methodName = logMethods[ii];
+                    log[methodName](methodName);
+                }
+
+                var passed = realLogMethod.calls.count() === expectedWorkingCalls;
```
pimterry: I think it's probably simpler and clearer to just put this inline as the value of 'pass' below.

pimterry:
Right, a couple of small comments there that we should solve first. On your other questions:

* I'm very happy to switch to Sinon, if you're happy to do that, that sounds great.

* I would rather not add a method to loglevel to get the current level, if possible (sorry if my previous comment was a bit ambiguous; I meant that the code generally looked fine). I don't think querying the current level is ever something that any real clients of the library should do in any sensible use case, so I'd like to avoid adding that just for the tests.

test/test-qunit.js
mroderick: OK, I can take a stab at doing that

PR #67, opened 2015-02-11
Explain change log level using localStorage method

medhajjej11:
Explain concretely how to change log level using the localStorage method

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ This is a barebones reliable everyday logging library. It does not do fancy thin
 ### Simple
 
 * Log things at a given level (trace/debug/info/warn/error) to the console object (as seen in all modern browsers & node.js)
-* Filter logging by level (all the above or 'silent'), so you can disable all but error logging in production, and then run log.setLevel("trace") in your console to turn it all back on for a furious debugging session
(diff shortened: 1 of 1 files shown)


pimterry:
In general you shouldn't need to mess with the local storage yourself directly, so I'm not keen to add this to the readme. What were you trying to do that meant you needed to change this manually?

madarche:
@pimterry I've been working with @medhajjej11 so I think I can anwser.

The use case was the following:

A company deploys a web app to production. This web app of course ships `loglevel`. By default no logs less important than `WARN` are outputted. Then one user has a problem and they call the support. The support needs the user to give more info about the problem. So the support asks the user to increase `loglevel` verbosity by writing `localStorage.loglevel='TRACE';` in the console.

But while writing this I realize that it's easier to have, let's say, a button on the UI that will programmatically switch the `loglevel` verbosity using `loglevel`'s API (the `log.setLevel(level)` method).

So this PR can be closed. Thanks for your kind listening.

pimterry:
Fair enough. Thanks for the extra detail @madarche. Yes, sounds like a button using this properly would be better, or direct use of the loglevel api (getting your customer to write `log.setLevel('info')`. I'll close this for now, let me know if you have any other thoughts on this, or how loglevel can support that better.

PR #72, opened 2015-05-28
Make log level persisting optional

BYK:
We want to set a higher log level in production by default but don't want it to be saved on users' browsers. Making persistence optional seemed like a good way to achieve this. The patch keeps the old behavior (persist by default) for backwards compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ The loglevel API is extremely minimal. All methods are available on the root log
   * As a string, like 'error' (case-insensitive) ← _for a reasonable practical balance_
   * As a numeric index from 0 (trace) to 5 (silent) ← _deliciously terse, and more easily programmable (...although, why?)_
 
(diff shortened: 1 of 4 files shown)


pimterry:
Thanks! In principle I'm fine with this, but could you please document it in the readme, and add a couple of tests covering it? Let me know if you need any help.

BYK:
bump :)

pimterry:
Ah, sorry, github doesn't ping when you add commits to a PR, only when you comment. Merged now, thanks for this!

BYK:
>Ah, sorry, github doesn't ping when you add commits to a PR, only when you comment.

No worries. I had the same issue so I should have pinged you earlier :)

Should I expect a new release on bower soon? ;)

pimterry:
New release (1.3.0) pushed, should be able to see it on Bower/NPM/etc immediately.

BYK:
Thank you!

Btw. sorry that I forgot the rebase, that ugliness, ugh :(

PR #73, opened 2015-06-03
Don't persist the default level

BYK:
Persisting the default level in production, especially when there was no
previous value persisted causes it being set on all clients for no
reason. It is also impossible to set a new default without somehow
cleaning up the old value.

### DIFF
--- a/lib/loglevel.js
+++ b/lib/loglevel.js
@@ -101,7 +101,7 @@
             storedLevel = "WARN";
         }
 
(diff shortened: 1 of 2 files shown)


pimterry:
Mmm, yes, sensible. Sorry for the delay. Merged!

pimterry:
and now released as 1.3.1

BYK:
>Sorry for the delay. Merged!

No worries. Yay! \o/

Issue #74, opened 2015-06-22
Ability to set level only if one is not persisted


Mr0grog:
In environments where a logging level has been explicitly set and persisted (usually a developer’s browser), it’d be nice to be able to leave the logging level alone, while otherwise giving a default logging level to use if one hasn’t been persisted. The only way to accomplish this right now is to explicitly peek into local storage or cookies to see if a value is set, which isn’t very nice.

I’m thinking this could either be a third argument to `setLevel` or, probably clearer, a separate method, e.g. `setDefaultLevel`.

Mr0grog:
I'm happy to do a PR for this if you'd like, too.

PR #75, opened 2015-06-25
Add a `setDefaultLevel()` method that sets levels only if none were previously persisted.

Mr0grog:
I had some time the other day, so I worked up this approach for solving #74. I realize it’s unsolicited, but hopefully it helps to have some working code to look at.

The idea is that, if your application automatically sets the logging level on startup, you’d use `setDefaultLevel()` instead of `setLevel()`. That way, if a developer or user opens the console and calls `setLevel()`, that setting will stick on the next page load (instead of being reset by app initialization). For example, your application might set the log level to `error` in a production environment, but when debugging an issue, you might call `setLevel("trace")` on the console to see all the logs. If that `error` setting was set using `setDefaultLevel()`, it will still say as `trace` on subsequent page loads and refreshes instead of resetting to `error`, which would have happened if page scripts used `setLevel()`.

I’m currently using a slightly more hacky version of this on a web site I’m working on (we check whether anything is stored under “loglevel” in local storage before calling `setLevel()`), but it would be nice to have it built in.

### DIFF
--- a/lib/loglevel.js
+++ b/lib/loglevel.js
@@ -84,7 +84,7 @@
         } catch (ignore) {}
     }
 
-    function loadPersistedLevel() {
+    function getPersistedLevel() {
         var storedLevel;
 
         try {
@@ -136,12 +136,9 @@
         }
     };
 
-    // NOTE: the semantics for the `persist` argument here are different than
-    // in `setLevel()`; it defaults to `false` instead of `true` (if you are
-    // setting a *default*, you likely don't want that persisted).
-    self.setDefaultLevel = function (level, persist) {
-        if (!loadPersistedLevel()) {
-            self.setLevel(level, !!persist);
+    self.setDefaultLevel = function (level) {
+        if (!getPersistedLevel()) {
+            self.setLevel(level, false);
         }
     };
 
@@ -164,6 +161,6 @@
         return self;
     };
 
-    self.setLevel(loadPersistedLevel() || "WARN", false);
+    self.setLevel(getPersistedLevel() || "WARN", false);
     return self;
 }));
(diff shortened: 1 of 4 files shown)


lib/loglevel.js
```diff
@@ -154,6 +164,6 @@
         return self;
     };
 
-    loadPersistedLevel();
+    self.setLevel(loadPersistedLevel() || "WARN", false);
```
pimterry: Can we rename this to getPersistedLevel, not load? Since the behaviour's now changed, and it's now just found and returned.

pimterry:
Sorry, I've been away for a little while, back now.

One small comment added on the code as it stands, but otherwise I'm happy with the feature in general. Thanks for following this up! The one specific thing I am less sure about is the persist argument for the default level. What's the use case for persisting a default level? Could we just make it always non-persistent instead, and keep things simpler?

Mr0grog:
> What's the use case for persisting a default level? Could we just make it always non-persistent instead, and keep things simpler?

Good point! I couldn’t think of any real strong case, which is why I flipped the default value. I left it in for parity with `setLevel()`, though I guess that’s kind of ruined when the default values differ anyway. You’re right; this would just be simpler without it. Will change.

lib/loglevel.js
Mr0grog: Can do.

Mr0grog:
Made all the changes you recommended, @pimterry. Looks like Travis failed because the Coveralls server threw an error when uploading results. The tests did pass, but I don’t have permission to re-run the build.

pimterry:
Looks good to me, merged. Thanks @Mr0grog! I'll put out a new release sometime soon.

Mr0grog:
Great! Happy to have helped :)

PR #77, opened 2015-07-08
Feature: Get current logging level

### LINKED ISSUE #70: Get current logging level.
For some operations it is good to know the current logging level.

For example, if you do string concatenation in your logging statements that are in sections of code that get called a lot, Javascript does the relatively-expensive string concatenation even if the current logging level won't actually output the string.

I think something like `log.getLevel()` would be useful so you could do the expensive operation only when needed.

madarche:
Fixes #70

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,21 @@ The loglevel API is extremely minimal. All methods are available on the root log
 
   These enable or disable all log messages, and are equivalent to log.setLevel("trace") and log.setLevel("silent") respectively.
 
+* A `log.getLevel()` method.
+
+  Returns the current logging level.
+  
+  For example, if you do string concatenation in your logging statements that
+  are in sections of code that get called a lot, Javascript does the
+  relatively-expensive string concatenation even if the current logging level
+  won't actually output the string.
+
+```javascript
+if (log.getLevel() >= log.levels.DEBUG) {
+    log.debug("Some " + "string " + "concatenation " + "called " + "a " + "lot");
+}
+```
+
 ## Plugins
 
 ### Existing plugins:
--- a/README.md
+++ b/README.md
@@ -150,9 +150,7 @@ The loglevel API is extremely minimal. All methods are available on the root log
           log.enableAll();
           msg = "Application in debug mode, read traces in the console.";
       }
-      p.innerHTML = '';
-      var p_text = document.createTextNode(msg);
-      p.appendChild(p_text);
+      p.innerHTML = msg;
   });
   ```
 
--- a/README.md
+++ b/README.md
@@ -128,17 +128,18 @@ The loglevel API is extremely minimal. All methods are available on the root log
 * A `log.getLevel()` method.
 
   Returns the current logging level.
-  
+
   For example, if you do string concatenation in your logging statements that
   are in sections of code that get called a lot, Javascript does the
   relatively-expensive string concatenation even if the current logging level
-  won't actually output the string.
+  won't actually output the string. So here is how to prevent the unnecessary
+  overhead through a condition on the current log level:
 
-```javascript
-if (log.getLevel() >= log.levels.DEBUG) {
+  ```javascript
+  if (log.getLevel() >= log.levels.DEBUG) {
     log.debug("Some " + "string " + "concatenation " + "called " + "a " + "lot");
-}
-```
+  }
+  ```
 
 ## Plugins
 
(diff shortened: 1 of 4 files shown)


madarche:
I've just added the documentation for this new API call.

madarche:
The proposed modification is really lightweight and, I believe, in the spirit of the loglevel call.

I agree that the previous code logic (before this PR) was smart by not storing the current level in memory. But being able to get the current log level is useful in some situations.

README.md
```diff
@@ -125,6 +125,22 @@ The loglevel API is extremely minimal. All methods are available on the root log
...
+  For example, if you do string concatenation in your logging statements that
+  are in sections of code that get called a lot, Javascript does the
+  relatively-expensive string concatenation even if the current logging level
+  won't actually output the string. So here is how to prevent the unnecessary
+  overhead through a condition on the current log level:
+
+  ```javascript
+  if (log.getLevel() >= log.levels.DEBUG) {
+    log.debug("Some " + "string " + "concatenation " + "called " + "a " + "lot");
+  }
```
pimterry: Can we come up with a better example here, with an actually expensive operation? This one isn't too great, because:

A) String concatenation is pretty cheap, and I'd be surprised if string concatenation in your logging is seriously the performance bottleneck in anybody's application. (Although if that is the case for you, I'd love to see the profiling data)
B) Even if this is the bottleneck, you can instead do `log.debug("Some ", "string ", "concatenation ", "called ", "a ", "lot")` (commas instead of pluses), and the console API will do concatenation for you, and automatically do this optimization too. It'll only concatenate the strings if debug logging is enabled, as if the level is higher than debug then log.debug is a no-op, so all the args are just thrown away.

I think this example needs to log something expensive, and the explanation shouldn't be about string concatenation either: it should be more general, and we should explicitly point out that most people won't need this, and you should only use it if they have measurements showing that logging is causing performance problems. I'm keen to avoid encouraging people into bad performance practices!

lib/loglevel.js
```diff
@@ -14,6 +14,7 @@
     }
 }(this, function () {
     var self = {};
+    var current_level;
```
pimterry: Convention everywhere else in this file is camelCase

test/default-level-test.js
```diff
@@ -54,6 +57,7 @@ define(['test/test-helpers'], function(testHelpers) {
                 log.setDefaultLevel("debug");
                 expect(log).toBeAtLevel("debug");
                 expect("debug").not.toBeTheStoredLevel();
+                expect(log.getLevel()).toBe(log.levels.DEBUG);
             });
         });
     });
```
pimterry: These shouldn't just be sprinkled into existing tests. Each test should be testing one specific thing, and each feature (including .getLevel()) should have standalone tests for whatever it's interesting cases are. Here if we accidentally break getLevel then a bunch of random unrelated tests fail, and it's not clear what's gone on and why.

Can you change this to add a couple of new tests for `getLevel()` instead, covering each of the relevant cases for getLevel()? I think those tests should probably be something like:

* Returns WARN initially
* Returns a new set value if setLevel is called
* Returns a new set value if setLevel is called even if it's not persisted

Can't think of any other cases, but feel free to add some if you can :-)

pimterry:
Few comments added, which I think need clearing up before we can merge this.

I'm still a little skeptical of this, as I'm a bit worried it's going to encourage people to do foolish things (like accidentally adding slightly different logic dependent on the log level, which is an easy way to shoot yourself in the foot really really badly when you want to debug things). I think the main thing is to make sure that the documentation makes it clear that this isn't something people should be using in 99% of cases, and it's just there for special cases and certain performance optimizations, if and when they're truly definitely required.

madarche:
@pimterry I totally agree with all your comments. I'll fix everything accordingly. It will take me a couple of days to find time to do this properly. Thanks Tim!

madarche:
@pimterry please could you check again now? Thanks.

pimterry:
I'm still not sure about the docs for this, but I can tweak that later after I've thought about what I'm aiming for, and that doesn't need to block the merge. Functionality all looks good to me, merged. Thanks for contributing!

madarche:
@pimterry thanks very much for the merge.

I hope you got the idea of our use case which is to display the current log level of a web app for the support team, so that they are able to toggle it.

And I'm watching the loglevel project and will be reading any change you will be doing on the doc with great interest. Thanks again for all your good work on loglevel!

PR #78, opened 2015-07-08
Add strict mode

madarche:
Because strict mode fixes mistakes that make it difficult for JavaScript
engines to perform optimizations: strict mode code can sometimes be made
to run faster than identical code that's not strict mode.

cf.  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode


pimterry:
Yep, sensible, merged. Thanks!

Issue #68, opened 2015-02-14
Supporting multiple loggers


rbabayoff:
Hey,

In large projects it is extremely useful to have the flexibility of defining different log levels for different modules / subsystems / files / objects, especially during debugging or bug chasing, in order to only see debug and trace log levels for a specific object / module. I have already implemented something like that in our wrapped loglevel meteor package:

https://github.com/practicalmeteor/meteor-loglevel

It also gave me the flexibility to define log prefixes that maintain line numbers (pushing the prefix as the first argument to apply), which, I suspect, is one of the more asked for features. 

Will you accept a PR that allows to create new loglevel() objects? Regarding persistence, the constructor can receive an options object that include a namespace string that will be used to load the log level for that namespace. It can also accept a prefix option and even a map of custom log levels to console log methods.

Mr0grog:
I’d definitely second this request. It could probably be done in a pretty light-weight fashion and would be extremely valuable for large projects, like one I’m consulting on now.

pimterry:
Great idea, I can definitely see how this could be useful. At the same time, I'm very much fighting to keep the loglevel API and codebase as simple as possible, since ultra-lightweight reliable quick easy logging is really the most important thing I'm going for here.

What would the API for this look like? If you can come up with something that's:

* Backward compatible
* Simple to use
* Simple to implement

Then I'd be happy for you to put a PR together. Definitely worth discussing the initial API before you do that though.

rbabayoff:
@pimterry, following is what I suggest as an API that is backwards compatible:

1. Create a new constructor function that actually creates and returns the logger.

2. The default global logger is automatically created on library load, as before.

3. To create a new logger, you call `log.newLogger()` or `log.createLogger()`, which calls the constructor function.

4. The constructor function supports 2 options, prefix and logLevel, both of which are optional and default to their current defaults - empty string, and WARN. I've already implemented this here:

https://github.com/practicalmeteor/meteor-loglevel/blob/master/loglevel-1.2.0.js#L3

5. I would also add a setPrefix method to the logger, as in here:

https://github.com/practicalmeteor/meteor-loglevel/blob/master/loglevel-1.2.0.js#L155

To improve on my implementation and to not possibly break code, I would only call the console functions with the prefix, if a prefix existed. 

6. Regarding persisting the log level, we can use the prefix (loglevel-prefix) to generate the key name for persistence.

[...]

PR #81, opened 2015-07-21
WIP: Multiple logger support

Mr0grog:
This is a simplistic implementation of #68 with minimal changes to the actual codebase (it looks a bit bigger than it really is because many methods moved into the logger constructor).

For the most part, all this ultimately does is add a method named `getLogger(name)` to the library. That method returns a new logger with the given name, or, if it’s already been called with that name, returns the previously created logger. The library itself is actually just an instance of one of these loggers, but it has the `getLogger()` and `noConflict()` methods sprinkled on top.

There are a couple ugly things in here, mostly pertaining to the discussion of inheritance: https://github.com/pimterry/loglevel/issues/68#issuecomment-120677667

- There’s a private variable named `defaultLoggerLevel` that tracks the current level of the default logger so it can be passed to new loggers. This would be better handled if logger instances had a `getLevel()` method, which would also solve #70 / fit with #77 

- There’s a little nuttiness with some private methods needing to be called with `.call()`. This is really because `methodFactory` needs to not be tightly bound to an instance in order for it to be passed along to new loggers. Maybe all private methods should be this way? Maybe some more rewriting would be in order, or a changed signature for `methodFactory`, or something related to having setters for `methodFacotry` (as noted in https://github.com/pimterry/loglevel/issues/70#issuecomment-121411810) would help. Haven’t thought too deeply here yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,13 +10,13 @@ Minimal lightweight simple logging for JavaScript. loglevel replaces console.log
 This is a barebones reliable everyday logging library. It does not do fancy things, it does not let you reconfigure appenders or add complex log filtering rules or boil tea (more's the pity), but it does have the all core functionality that you actually use:
 
 ## Features
- 
+
 ### Simple
 
 * Log things at a given level (trace/debug/info/warn/error) to the console object (as seen in all modern browsers & node.js)
 * Filter logging by level (all the above or 'silent'), so you can disable all but error logging in production, and then run log.setLevel("trace") in your console to turn it all back on for a furious debugging session
 * Single file, no dependencies, weighs in at <1KB minified and gzipped
- 
+
 ### Effective
 
 * Log methods gracefully fall back to simpler console logging methods if more specific ones aren't available: so calls to log.debug() go to console.debug() if possible, or console.log() if not
@@ -110,9 +110,9 @@ The loglevel API is extremely minimal. All methods are available on the root log
   * As a log level from the internal levels list, e.g. log.levels.SILENT ← _for type safety_
   * As a string, like 'error' (case-insensitive) ← _for a reasonable practical balance_
   * As a numeric index from 0 (trace) to 5 (silent) ← _deliciously terse, and more easily programmable (...although, why?)_
-  
+
   Where possible the log level will be persisted. LocalStorage will be used if available, falling back to cookies if not. If neither is available in the current environment (i.e. in Node), or if you pass `false` as the optional 'persist' second argument, persistence will be skipped.
-  
+
   If log.setLevel() is called when a console object is not available (in IE 8 or 9 before the developer tools have been opened, for example) logging will remain silent until the console becomes available, and then begin logging at the requested level.
 
 * A `log.setDefaultLevel(level)` method.
@@ -120,7 +120,7 @@ The loglevel API is extremely minimal. All methods are available on the root log
   This sets the current log level only if one has not been persisted and can’t be loaded. This is useful when initializing scripts; if a developer or user has previously called `setLevel()`, this won’t alter their settings. For example, your application might set the log level to `error` in a production environment, but when debugging an issue, you might call `setLevel("trace")` on the console to see all the logs. If that `error` setting was set using `setDefaultLevel()`, it will still say as `trace` on subsequent page loads and refreshes instead of resetting to `error`.
 
   The `level` argument takes is the same values that you might pass to `setLevel()`. Levels set using `setDefaultLevel()` never persist to subsequent page loads.
-  
+
 * `log.enableAll()` and `log.disableAll()` methods.
 
   These enable or disable all log messages, and are equivalent to log.setLevel("trace") and log.setLevel("silent") respectively.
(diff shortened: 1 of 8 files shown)


test/multiple-logger-test.js
```diff
@@ -0,0 +1,84 @@
...
+            });
+
+            it("loggers are created with the same methodFactory as the default logger", function(log) {
+                log.methodFactory = function(methodName, level) {
+                  return function() {};
+                };
+
+                var newLogger = log.getLogger("newLogger");
+                expect(newLogger.methodFactory).toEqual(log.methodFactory);
+            });
```
pimterry: Would be good to include a test checking that it picks up the persisted level for this logger, if that is already set, and doesn't inherit in that case.

And a test checking that I can get a new logger, set the default logger level, get the new logger again, and the log level hasn't changed (i.e. test it only inherits from the default logger when it's first created, and then the log level is totally independent).

test/multiple-logger-test.js
```diff
@@ -0,0 +1,84 @@
...
+            it("should throw if called with no name", function(log) {
+                expect(function() {
+                  log.getLogger();
+                }).toThrow();
+            });
+
+            it("should throw if called with a non-string name", function(log) {
+                expect(function() {
+                  log.getLogger(true);
+                }).toThrow();
```
pimterry: Should also cover the empty string case.

test/multiple-logger-test.js
```diff
@@ -0,0 +1,84 @@
...
+                var newLogger = log.getLogger("newLogger");
+                expect(newLogger.getLogger).toBeUndefined();
+                expect(newLogger.noConflict).toBeUndefined();
+            });
+
+            it("returns the same instance when called repeatedly with the same name", function(log) {
+                var logger1 = log.getLogger("newLogger");
+                var logger2 = log.getLogger("newLogger");
+
+                console.log("Logger: ", logger1);
```
pimterry: Is there a reason this is in here? Looks like leftover debugging, should probably go.

lib/loglevel.js
```diff
@@ -47,122 +46,169 @@
...
-        }
-    };
 
-    self.setDefaultLevel = function (level) {
-        if (!getPersistedLevel()) {
-            self.setLevel(level, false);
+        var logger = _loggersByName[name];
+        if (!logger) {
+          logger = _loggersByName[name] = new Logger(
+            name, defaultLoggerLevel, defaultLogger.methodFactory);
```
pimterry: This is a nice solution to the default level I think, good stuff. As you note though, it'd be good to remove defaultLoggerLevel, and `getLevel()` is merged now so if you rebase it should be easy to strip that out.

lib/loglevel.js
```diff
@@ -47,122 +46,169 @@
...
-               enableLoggingWhenConsoleArrives(methodName, level);
-    };
+    var defaultLogger = new Logger();
 
-    self.setLevel = function (level, persist) {
-        if (typeof level === "string" && self.levels[level.toUpperCase()] !== undefined) {
-            level = self.levels[level.toUpperCase()];
+    var _loggersByName = {};
+    defaultLogger.getLogger = function getLogger(name) {
+        if (!name || typeof name !== "string") {
```
pimterry: The `!name` case isn't needed for null/undefined/etc, since those aren't strings anyway. This is only catching the empty string case. It'd be nicer to do that explicitly, so it's easy to see that that's not an allowed value without having to think about how strings coerce to booleans:

```javascript
if (typeof name !== "string" || name === "") { ... }
```

lib/loglevel.js
```diff
@@ -47,122 +46,169 @@
...
+        "debug",
+        "info",
+        "warn",
+        "error"
+    ];
 
-        return storedLevel;
+    var defaultLoggerLevel;
+
+    function Logger(name, level, factory) {
```
pimterry: I'd call this 'defaultLevel', since that's what it is. It's not necessarily the actual level of the logger, as it gets overridden by the persisted level if one exists.

pimterry:
Great work, thanks for chasing this! I think overall this approach is very good. I've put a couple of comments inline on specific tweaks, and there's a couple of wider things:

* This needs rebasing, as I've now merged the `getLevel()` PR.

* Needs documentation.

* I do think the code would be much nicer if we didn't need the `.call(this)` stuff. I don't have an easy answer I'm afraid. Moving the logger parameter from being a `this` that we have to remember to instead being an explicit argument would be an easy improvement on the `enableLoggingWhenConsoleArrives` and `replaceLoggingMethods` methods though.

  The bit where that becomes tricky is the methodFactory method itself, which I'm not as sure how to handle. Maybe that should just take a logger argument too, but that's not something most plugins should need to implement a method factory, and I'm worried it'll encourage to people build plugins badly (e.g. checking logger.getLevel() to decide whether the log factory should be building a method, rather than just leaving that to the logger and focusing on returning a new method). Maybe that's not something to be concerned about though. I don't have any easy answers there, any more thoughts you had would be great.

* Method factory more generally: I am currently planning to move to a setMethodFactory() API very soon. I'm either going to deprecate the previous approach and support both for a little while, or just bite the bullet and release all this as v2.0 instead. Just FYI, in case that helps your approach to the previous point.

test/multiple-logger-test.js
Mr0grog: D’oh! That’s exactly what this is. Thanks for catching it.

test/multiple-logger-test.js
Mr0grog: > test it only inherits from the default logger when it's first created, and then the log level is totally independent

Heh, I think I was trying to still leave that issue open for discussion, but that was also probably just wishful, rather than realistic thinking on my part. Will add.

Mr0grog:
Ok, I spent a little time this afternoon cleaning up the tests, adding docs, and fixing `methodFactory` so it gets the right args (forgot about that last time around).

I *haven’t* worked on better solutions to the private-but-shared-across-instances issue with `enableLoggingWhenConsoleArrives` et al. However, some thoughts:

> Maybe that should just take a logger argument too, but that's not something most plugins should need to implement a method factory

I’m skeptical that a third-party plug-in shouldn’t be capable of doing something the built-in one can. Besides, the current implementation already sends `methodFactory` the logger—just as `this` rather than an argument.

That said, I get the idea that the API should hopefully push developers towards doing the “right” thing. Maybe one approach would be to send the methodFactory a function that resets the logging methods (i.e. a version of `replaceLoggingMethods` that takes no arguments) and *not* be given the actual logger instance as `this`. That severely constrains what a plug-in can do, while still enabling current functionality. But maybe all this is better discussed in relation to v2 or adding `setMethodFactory()`. Is there an issue for that?

README.md
```diff
@@ -130,21 +130,63 @@ The loglevel API is extremely minimal. All methods are available on the root log
...
+  var moduleTwo = require("module-two");
+  log.getLogger("module-two").setLevel("TRACE");
+
+  moduleOne.doSomethingAmazing();
+  moduleTwo.doSomethingSpecial();
+  // logs "Special message from module two."
+  // (but nothing from module one.)
+  ```
+
+  Loggers returned by `getLogger()` support all the same properties and methods and default, root logger, excepting `noConflict()` and the `getLogger()` method itself.
```
pimterry: 'and default, root logger' -> 'as the default root logger'

README.md
```diff
@@ -155,22 +197,22 @@ ServerSend - https://github.com/artemyarulin/loglevel-serverSend - Forward your
...
+There's clearly enough enthusiasm for this even at that cost though that loglevel now includes a plugin API. To use it, redefine log.methodFactory(methodName, logLevel, loggerName) with a function of your own. This will be called for each enabled method each time the level is set (including initially), and should return a function to be used for the given log method, at the given level, for a logger with the given name. If you'd like to retain all the reliability and features of loglevel, it's recommended that this wraps the initially provided value of `log.methodFactory`
 
 For example, a plugin to prefix all log messages with "Newsflash: " would look like:
 
 ```javascript
 var originalFactory = log.methodFactory;
-log.methodFactory = function (methodName, logLevel) {
+log.methodFactory = function (methodName, logLevel, loggerName) {
     var rawMethod = originalFactory(methodName, logLevel);
```
pimterry: This example won't actually work any more: the method factory needs to be called with the right context :smile:. Hopefully we'll find a better solution there, so no need to fix this for now.

Mr0grog: Ha, good point. I guess I should really make sure this example continues to work, though. Which will unfortunately be kind of ugly, since we’d need to keep track of the unbound version so we can re-bind it to non-default loggers. Also it looks like I forgot to add the third argument to the `originalFactory()` call :\

Mr0grog: (Or I should change the example to `originalFactory.call()`, but it seems better to make it continue to work as it has.)

pimterry:
> I’m skeptical that a third-party plug-in shouldn’t be capable of doing something the built-in one can. Besides, the current implementation already sends methodFactory the logger—just as this rather than an argument.
> That said, I get the idea that the API should hopefully push developers towards doing the “right” thing. 

Mm, my concern isn't really so much what's possible, but what the API encourages people to do. If you only need the logger in strange and wonderful cases, I'd rather not pass it as an explicit argument. Providing it only as the context feels less visible and suggestive, although I realise that's a pretty fluffy argument.

> But maybe all this is better discussed in relation to v2 or adding setMethodFactory(). Is there an issue for that?

There is now: #82.

I was thinking this wouldn't make much difference, but actually I think it really does. A big part of this current problem is that the current method factory API is quite unfriendly to this, and we're exposing the default factory, so we have to be careful to make that usable. If we're breaking the API anyway we can stop doing both of those though, so that writing a plugin looks more like:

```javascript
log.wrapLogMethods(function (defaultLogMethod, methodName, loggerName) {
  return function (message) {
    defaultLogMethod("Newsflash: ", message);
  };
});
```

No need to read the current log factory (which is good, because that field being exposed is part of why I want to change this), and nobody ever sees the default method factory at all, only its output.

That's all great for this, as it turns out, because that means we can split the internal representation. Rather than just having a `log.methodFactory` property for everything, we can have a private methodFactory function inside the Logger (lexically inside, so it can closure in `self` like every other private function), plus a separate customMethodWrapper field, which is undefined unless set with `.wrapLogMethods`, and which can be safely copied between instances.

Some of those details are still a bit vague, but I'm reasonably confident that this means we can clean this up easily once that's added. Given that, lets just ignore this for now.

Other than things the next change will fix I think this looks great, so I'm going to merge it now and then get started on the improved plugin API. Any thoughts you have on all this would be very interesting though, over on #82.

Thanks!

davetayls:
looking forward to this one :+1:

PR #83, opened 2015-08-04
Updated bower main entry for pre-release version

andviro:
Hi! Thanks for creating this, excellent logging library. I've noticed that when using bower for package management I lose latest features like getLogger(), hence this small patch.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "loglevel",
   "version": "1.3.1",
-  "main": "dist/loglevel.min.js",
+  "main": "lib/loglevel",
   "dependencies": {},
   "ignore": [
     "**/.*",


pimterry:
Ah, I see. This needs to be pointing at the built dist version, not the lib, because that's the official known-good version, and if I update the bower.json I'll have to change it back every time we do build a real release, so I'm not going to merge this I'm afraid.

If you use the /lib/ version you've got no guarantees it won't suddenly change, and it's code that hasn't been properly tested, and might contain all sorts of backward incompatible crazy changes. You're welcome to use it anyway, but it may fall apart for you :-).

Anyway, the problem here really is that the bower config is only intended for official releases, not prerelease use. If you want to get the prerelease features, I'd suggest you do one of:

* Download the prerelease version manually, drop it into your project and then upgrade to the release when it lands soon separately
* Depend on your prerelease-configured fork through bower, and then upgrade to the release when it lands soon separately
* Muddle through without for now, and wait a week or two until I push a new official release with all these features in and everything magically works for you

Sorry about that, thanks for contributing anyway!

andviro:
Yeah, sorry for the fuss. Thanks for excellent library, I'll be waiting for the next npm release.

PR #92, opened 2016-05-18
Switch order of module pattern checks for more consistent UMD.

kaidjohnson:
I was having errors thrown bundling loglevel in a library that was then being bundled in an application package and found that reversing these checks solves the issue for us.

Based on this conversation: https://github.com/umdjs/umd/pull/22 it seems that checking for AMD before CommonJS is a safer paradigm to follow.

### DIFF
--- a/lib/loglevel.js
+++ b/lib/loglevel.js
@@ -7,7 +7,7 @@
 (function (root, definition) {
     "use strict";
     if (typeof define === 'function' && define.amd) {
-        define('loglevel',definition);
+        define(definition);
     } else if (typeof module === 'object' && module.exports) {
         module.exports = definition();
     } else {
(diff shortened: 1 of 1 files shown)


pimterry:
That sounds good, but this change has broken all the tests. Can you take a look at why? I'm happy to merge a change like this, just as long as the tests pass :smile:

kaidjohnson:
Hey - that's because I'm really good at breaking tests inadvertently. Apparently, my copy-paste skills are rusty and my amd definition did not properly parallel what already existed in code. 

I've updated and local tests went from red to green, so with a little bit of luck, the travis ci build should do the same.

pimterry:
Looks great, thanks! Merged.

JoHense:
Thanks for this fix!
@pimterry do you plan do a minor release in the near feature? I would like to use loglevel with AMD as soon as possible :-)

pimterry:
All good now; this is now up and released as 1.4.1. Let me know if you have any other trouble.

Issue #93, opened 2016-05-26
Extend the API with something like `loglevel.getLoggers()`


myme:
Since `loglevel` supports sub-loggers through the `loglevel.getLogger()` function, would it be acceptable to have a function for listing existing loggers? I see that the object containing the logger references is within the module's closure and not accessible from the outside (which is good imho).

However, I'm wondering if it would be nice to have a function that reads the keys of this object and returns them as a list perhaps? Or some other structure which makes sense. I think this could be really useful when debugging an application which dynamically creates loggers, or if you have *many* loggers.

pimterry:
As discussed in #94: this is implementable as a standalone plugin and (until there are common use cases where it would be particularly useful) that's where it should live for now.

PR #94, opened 2016-05-26
Add `log.getLoggers()`

myme:
New API method for returning a list of all the named loggers created
with `log.getLogger()`.

This may be useful if the number of sub-loggers grow very large or if
the application dynamically creates loggers.

This pull request is a simple implementation of #93. I've not read up on the compatibility goals for loglevel, but `Object.keys` can easily be exchanged for something else for better compatibility across older interpreter implementations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,21 @@ The loglevel API is extremely minimal. All methods are available on the root log
 
   Likewise, loggers will inherit the root logger’s `methodFactory`. After creation, each logger can have its `methodFactory` independently set. See the *plugins* section below for more about `methodFactory`.
 
+* A `log.getLoggers()` method.
+
+  This returns a list of all the named loggers which have been created with `log.getLogger()`. The root logger is not included in this list.
+
+  ```javascript
+  var log = require("loglevel");
+
+  // Create a few loggers
+  log.getLogger("logger1");
+  log.getLogger("logger2");
+
+  // Returns '["logger1", "logger2"]'
+  log.getLoggers();
+  ```
+
 
 ## Plugins
 
--- a/lib/loglevel.js
+++ b/lib/loglevel.js
@@ -208,6 +208,10 @@
         return logger;
     };
 
+    defaultLogger.getLoggers = function getLoggers() {
+      return Object.keys(_loggersByName);
+    };
+
     // Grab the current global log variable in case of overwrite
     var _log = (typeof window !== undefinedType) ? window.log : undefined;
     defaultLogger.noConflict = function() {
--- a/test/multiple-logger-test.js
+++ b/test/multiple-logger-test.js
@@ -63,6 +63,21 @@ define(['test/test-helpers'], function(testHelpers) {
             });
         });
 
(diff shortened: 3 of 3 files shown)


pimterry:
Interesting, ok.

I'm still not clear on the use case for this. Is there somewhere in your codebase where you would actually use this? I'm not strongly against it, but I'd rather not increase the complexity of loglevel without concrete benefits.

Compatibility is also a concern: the overall goal of loglevel is that it fails *nowhere* (realistically, in IE6+). It doesn't matter if features don't do anything in environments where they're not available, but it should make it easy to use it to write logging code that works happily in nice environments, has sensible workarounds for less good environments (IE 8 + 9, for example), and doesn't break elsewhere. It's not as rigorously enforced as I'd like sadly (the testing libraries don't actually work in IE6), but I do some quick manual tests for core functionality in as many old browsers as I can when I release.

Anyway, yes, in practice I think this just means changing this away from Object.keys, to something simpler that will work sensibly everywhere.

myme:
I didn't necessarily expect this PR to be merged as-is, but I wanted the feature request to have a proof of concept.

Regarding the use case and my own codebase: I was thinking about making a simple log management HUD for a webapp, where the developer can go to toggle the debug level of the different named loggers (and the root logger of course). In order to do this nicely, the HUD should be populated programmatically by reading out the list of active loggers.

I haven't made the UI for it yet, but it was quite easy to add this feature through a plugin (which also prefixes named loggers), and use the browser's console to list and fiddle with loggers:

```javascript
import log from 'loglevel';

const _origMethodFactory = log.methodFactory;
const _loggers = new Set();

export default Object.assign(log, {
  methodFactory(methodName, logLevel, loggerName) {
    if (loggerName) {
      _loggers.add(loggerName);
    }
    const rawMethod = _origMethodFactory(methodName, logLevel, loggerName);
    return (...args) => {
      rawMethod(`[${loggerName || 'root'}]`, ...args);
    };
  },

  getLoggers() {
    return [..._loggers];
  },
});

log.setLevel(log.getLevel());
```

I understand that this may not be mainstream enough for it to justify an update of the `loglevel` API, but it could spark a discussion. Like I said, it's quite easy to create a plugin for this, so having this in doesn't block me at all. I was just thinking that it may be of interest to more users of the tool.

pimterry:
Ok, that makes sense. Building a logger HUD isn't really standard usage (although that does sound cool), and given that you can implement it in a plugin nicely, I think that's probably where this should go. I'm very keen to keep the core library small, clean and focused.

I'll close this PR for now (but thanks for getting involved anyway! Definitely worth discussing).

If you do release your plugin code standalone, feel free to open a PR to add it to the list of plugins in the README, I'd be very happy to link to it.

myme:
Cool. I respect the fact that you want to keep the `loglevel` implementation at a minimum. In any case, it's nice to have these discussions and "solutions" laying around for others to stumble upon. I'll surely submit a PR should the plugin ever find its way to the public domain. Thanks.

PR #104, opened 2017-05-11
Check if window is available before using it

thomasconner:
Some JavaScript platforms break because of the use of window. Below are some platforms I can think of.

https://www.nativescript.org/
http://www.appcelerator.com/mobile-app-development-products/
https://facebook.github.io/react-native/ (Possibly)

### DIFF
--- a/lib/loglevel.js
+++ b/lib/loglevel.js
@@ -101,40 +101,40 @@
       function persistLevelIfPossible(levelNum) {
           var levelName = (logMethods[levelNum] || 'silent').toUpperCase();
 
-          if (typeof window !== undefinedType) {
-            // Use localStorage if available
-            try {
-                window.localStorage[storageKey] = levelName;
-                return;
-            } catch (ignore) {}
+          if (typeof window !== undefinedType) return;
 
-            // Use session cookie as fallback
-            try {
-                window.document.cookie =
-                  encodeURIComponent(storageKey) + "=" + levelName + ";";
-            } catch (ignore) {}
-          }
+          // Use localStorage if available
+          try {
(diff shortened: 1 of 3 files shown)


lib/loglevel.js
```diff
@@ -101,36 +101,40 @@
...
+                window.localStorage[storageKey] = levelName;
+                return;
+            } catch (ignore) {}
+
+            // Use session cookie as fallback
+            try {
+                window.document.cookie =
+                  encodeURIComponent(storageKey) + "=" + levelName + ";";
+            } catch (ignore) {}
+          }
```
pimterry: Sure, this is reasonable. I'd rather flip it and add a guard clause though, rather than indenting the whole function. Can you convert both of these into something like the below instead?

```javascript
if (typeof window === undefinedType) return;

[...rest of the function...]
```

thomasconner: Sure

pimterry:
Sorry, JSHint actually doesn't like that example, so the build broke. I disagree though, so I just went and added a quick patch to relax that, and then there was an actual bug, so I fixed that while I was there too. Anyway, all looks good to me now, merging! Thanks for contributing :smiley:

pimterry:
@thomasconner Hmm, wait actually. Looking closer at this, I'm not sure why it's required. The following lines that access `window` are all wrapped with a try/catch, so everything should still work fine without `window` present anyway.

Are you fixing this because this causes an actual issue you can reproduce somewhere?

thomasconner:
Yes. When I was using `loglevel` in a NativeScript app it would crash because it was trying to access `window`. 

I will try and put a project that reproduces the issue.

Issue #100, opened 2016-12-23
Add ES6/browserify/vuejs docs to README?


OmgImAlexis:
Since more and more people are starting to use Vuejs it may be an idea to add something to the docs to let them know how to get this library working.

```
import * as log from 'loglevel';
window.log = log.noConflict();
```

pimterry:
I don't know vue. I'm surprised by the above though - why do you need to do that, why don't you just import log and then use it like any other library?

OmgImAlexis:
You need to import all the functions otherwise none of the log.* functions works.
For example `log.info()` only works after importing info from `loglevel`.

You also need window.log to use it within all of your .vue files otherwise it'll only work in the single file/component that it's imported into. If the user wanted they could just import it and bind `log.noConlict()` to the Vue method field but that would limit log to only be usable for that single component.

mibanez:
Still not working for me (i'm using `vue-cli`). Seems to be that `loglevel` has not access to console and every log method end up being the default method `noop` (function() {})

OmgImAlexis:
@mibanez did you use `import * as log from 'loglevel';`? If you just import loglevel itself then nothing it set, that's why you end up with noop.

julestruong:
@OmgImAlexis not working either for me, did you actually test your code ? :/

OmgImAlexis:
Yes and I use it for multiple projects.

julestruong:
@OmgImAlexis worked , sorry

pimterry:
Glad to hear that all worked out. I'm not totally sure I'm following the above as a non-vue user. Are there some basic instructions for this vue setup that should be in the readme? If any of you want to put together a PR to add a short snippet for this (under [`Setting it up`](https://github.com/pimterry/loglevel#setting-it-up)), I'd be happy to include it.

OmgImAlexis:
@pimterry just including something like this would work; and it's for ES6 not vue specifically.

In ES6 make sure to import the whole package.
The last line only applies while using ES6 in the browser as allows the use of log anywhere in your app.
```
import * as log from 'loglevel';
window.log = log.noConflict();
```

PR #107, opened 2017-07-18
Change ES6 import in README

danrot:
I would propose to change the es6 import as done in this PR. It practically works exactly the same then, except for one crucial difference: If imported this way the references will be updated when changing the log level using `setLevel` at runtime. It took me some time to realize that the way the methods have been imported were the issue. Using `*` seems to be resulting in a separate reference, which doesn't care about the [`replaceLoggingMethods` method](https://github.com/pimterry/loglevel/blob/master/lib/loglevel.js#L68).

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ log.error("too easy");
 ### As an ES6 module (assuming some transpilation step):
 
 ```javascript
-import * as log from 'loglevel';
+import log from 'loglevel';
 log.debug("all done");
 ```
 


pimterry:
I think this is something that depends a lot on the environment in your project.

In my experience, I've usually seen ES2015 modules used with a compiler that simply used the CommonJS/AMD output of the module as the full module object. That means `import log from 'loglevel'` is equivalent to `var log = require('loglevel').log` in CommonJS land, and that's code that doesn't work (there's no such property), so I'm pretty surprised it's working for you.

The current code (`import * as log from 'loglevel'`) should typically be equivalent to `var log = require('loglevel')`, which is correct. This was reported by a bunch of Vue users, who all seemed to need this form: https://github.com/pimterry/loglevel/issues/100

There's obviously other ways to handle this though, and I'm not enormously well-versed all in the many ways you can use ES2015 modules in practice. Can you give me some more detail on how you're building your project?

danrot:
Since it is also an open source project I can share it with you: https://github.com/sulu/sulu-minimal/blob/develop/webpack.config.js

Basically we are using webpack in combination with babel.

pimterry:
> `import log from 'loglevel'` should import the default export 

Doh, yes, of course: I was thinking of `import { log } ...`.

This module doesn't have a default export of course though - there's no `.default` property on the result at all. It looks like that's a Babel-specific solution to this (http://2ality.com/2015/12/babel-commonjs.html#how-babel-imports-commonjs-modules) (where `.default` instead imports the whole object). It's certainly different to however the originating Vue developers' compilation worked though, and how I've seen it work in ES2015 module implementations elsewhere (TypeScript definitely doesn't do this, for example). 

As far as I can tell there isn't really an officially 'right' answer to this yet, because Node hasn't settled on exactly how its internal commonjs to ES2015 mapping is going to work either. I'd be happy to accept a change that included your version here as an alternative though somehow, e.g. 'for babel commonjs module compilation, ...' and 'for some other ES2015 module implementations, including TypeScript, ...'

I'm still a bit confused by your initial comment: 'if imported this way the references will be updated when changing the log level using setLevel at runtime'. The exported `log` object that `import * as log ...` imports never gets changed, it's just its properties that change, but that should be fine. Any chance you were using `import { error, warn, ... } from 'loglevel'`? That would definitely break as you describe.

killercentury:
I do think we need to add some more clarification on this ES6 usage. Like the React community is heavily utilizing ES6 and Babel, and they will be going to look at this docs as well. But when they do `import * as log from 'loglevel'`, what happens is that there is two log objects ("log" and "log.default") are actually being created (refer to http://2ality.com/2015/12/babel-commonjs.html). While "log.default" is original object, if we do `log.setLevel`, "log.default" will be updated properly, but "log" will stay unchanged.

danrot:
@pimterry The properties actually don't change (as also @killercentury described), that's the exact problem :-)

pimterry:
Just to reiterate my comment above: while I don't think this current update is ok, as it won't work for everybody, I'm very open to a similar change to fix this, if somebody wants to do some research/testing and come up with a more general version, and a broad outline of who the different forms are relevant to. E.g. 'for babel commonjs module compilation, ...' and 'for some other ES2015 module implementations, including TypeScript, ...'

pimterry:
No longer relevant - this is now fully documented by https://github.com/pimterry/loglevel/commit/1ee5d694bc5ccd7045c2fa357d8f55ec5b0723f5

PR #112, opened 2017-10-08
Fixed cookie test.

quotquot:
indexOf returns -1 when the item is not found.


pimterry:
Great find, thanks! Merged and released as 1.5.1.

PR #114, opened 2017-11-02
Add getLoggers method and save name of loggers on instance

iabw:
Hi! Tidy little lib, but I wanted to be able to generate a list of all my loggers so the user can select the level for each logger themselves. This was all I needed to do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,9 @@ The loglevel API is extremely minimal. All methods are available on the root log
 
   Likewise, loggers will inherit the root logger’s `methodFactory`. After creation, each logger can have its `methodFactory` independently set. See the *plugins* section below for more about `methodFactory`.
 
+* A `log.getLoggers()` method.
+
+  This will return you the dictionary of all loggers created with `getLogger`, keyed off of their names.
 
 ## Plugins
 
--- a/lib/loglevel.js
+++ b/lib/loglevel.js
@@ -158,6 +158,8 @@
        *
        */
 
+      self.name = name;
+
       self.levels = { "TRACE": 0, "DEBUG": 1, "INFO": 2, "WARN": 3,
           "ERROR": 4, "SILENT": 5};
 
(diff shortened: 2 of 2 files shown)


pimterry:
Looks great, merged. Thanks!

pimterry:
@iabw that was already possible - that code could just register a plugin, which would get called any time any logger was registered and given its name.

In general, loglevel doesn't provide any privacy within the same JS process, and I'm not sure any standalone JS library really can at all without reimplementing an entire JS runtime. That's not really an existing feature of the language. By the time you've got untrusted code running inside your javascript process, everything else in that process is effectively public, since there's no real true isolation or inaccessibility mechanisms in JS. Untrusted JS could read the source of the whole app and parse out all the logger names by hand to do this, for example, or even transform and then rerun loglevel's initialization code to make private vars public for itself.

Even the [VM module](https://nodejs.org/api/vm.html) in node, which runs code in a whole new V8 virtual machine has a big warning says:

> Note: The vm module is not a security mechanism. **Do not use it to run untrusted code.**

There are specialized isolation mechanisms that try to do this kind of thing, like [Jailed](https://github.com/asvd/jailed) and [Google Caja](https://developers.google.com/caja/) (though I've never tried either), and the problem as a whole is the subject of some interesting academic research: https://seclab.stanford.edu/websec/jsPapers/csf09-camera-ready.pdf.

I've never seriously looked into it myself, but you're definitely not going to get any serious isolation from almost any random JS library - if you truly want to run untrusted JS, it's going to be a major piece of separate isolation development that will significantly change the architecture of your code, or it's going to be broken. I'd recommend just not doing it :smiley:.

iabw:
That's totally fair, I guess I meant more of either a footgun or "unlocked doors" vs "break in through the window". You've obviously put a lot of thought into this, so if you're cool with the getter, I'm happy to use it.

PR #121, opened 2018-06-18
Add note about Chrome log filtering

RichardBradley:
It took me a while to figure out why my `debug` logs weren't showing, perhaps this will help other users?

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,10 @@ The loglevel API is extremely minimal. All methods are available on the root log
 
   These methods should never fail in any environment, even if no console object is currently available, and should always fall back to an available log method even if the specific method called (e.g. warn) isn't available.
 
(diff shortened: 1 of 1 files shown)


kutuluk:
loglevel uses console.log instead of console.debug and does not require any additional settings.

RichardBradley:
These settings were necessary for me to have logging sent from `loglevel.debug("...")` to appear in Chrome ver 67.0.3396.79 on Win 10

pimterry:
@RichardBradley which version of loglevel are you using? Since 1.5.0 `loglevel.debug` should map directly to console.log. If you're using that or later, and debug messages don't show up by default, that sounds like a bug.

RichardBradley:
I'm on "1.4.1", yes, sorry. I'll update.

pimterry:
Sure, no worries! Glad you got it sorted 👍

PR #131, opened 2019-05-13
fix variadic typo

0xflotus:


pimterry:
Nice spot! Thanks :+1:

PR #143, opened 2020-04-13
Add logger.log() type

KishanRPatel:
I am using `loglevel` in a project with TypeScript, and wanted to migrate my `console.log` calls to `log.log` calls. It didn't compile because the `Logger` interface was missing the `log` function type. I added it, and it fixed my issue.


pimterry:
Looks good, thanks! Merged :+1:

PR #144, opened 2020-04-14
Fix typo in README

ericboucher:
still *say* as `trace` -> still stay as `trace`


pimterry:
Good spot, thanks! :+1:

PR #152, opened 2020-09-02
docs: Fix simple typo, blantantly -> blatantly

timgates42:
There is a small typo in test/test-qunit.js.

Should read `blatantly` rather than `blantantly`.


pimterry:
Nice catch, thanks!

PR #167, opened 2021-05-06
Update README.md improve Writing plugins

### LINKED ISSUE #166: methodFactory logLevel parametr has always same value
When I set my own `loglevel.methodFactory` then the parameter has always same value - the min. number set by `setLevel`.
E.g. https://jsfiddle.net/xmedeko/e9z1038x/9/ :
```js
var originalFactory = log.methodFactory;
log.methodFactory = function (methodName, logLevel, loggerName) {
  var rawMethod = originalFactory(methodName, logLevel, loggerName);
  return function (...args) {
    rawMethod(methodName, logLevel, ...args);
  }
}

log.setLevel(1);
log.info("info message");
log.warn("warn message");
log.error("error message");
```
The output shows `logLevel` param as 1, 1, 1. Is this behaviour ok?

I would expect it to be 2, 3, 4 according to the `methodName`. So as I can do e.g.
```js
log.methodFactory = function (methodName, logLevel, loggerName) {
  var rawMethod = originalFactory(methodName, logLevel, loggerName);
  return function (...args) {
    rawMethod(...args);
    if (logLevel >= 3) sendLogSomewhere(...);
  }
}
```

xmedeko:
Fix #166


pimterry:
Nice improvement, thanks! :+1:

PR #168, opened 2021-06-16
[Feature]: Add formatted Suffix

itamayo:
* Modification in order to add suffix to logging without loss console log main functionality such as tracing:

Basic example:

```js

const  socketLog = logger.getLogger('socketLog');
socketLog.setPrefix("[SocketLog]");
socketLog.info("test");

```
Formating text:

```js
const  socketLog = logger.getLogger('socketLog');
chicken.setPrefix(formating.bind(socketLog,"socketLog"));
socketLog.info("test");
function formating(name){ 
   return  new Date().toTimeString().replace(/.*(\d{2}:\d{2}:\d{2}).*/, '$1') + `(${name || this.name})`
}
```

### DIFF
--- a/lib/loglevel.js
+++ b/lib/loglevel.js
@@ -1,9 +1,4 @@
-/*
-* loglevel - https://github.com/pimterry/loglevel
-*
-* Copyright (c) 2013 Tim Perry
-* Licensed under the MIT license.
-*/
+/*! loglevel - v1.7.1 - https://github.com/pimterry/loglevel - (c) 2020 Tim Perry - licensed MIT */
 (function (root, definition) {
     "use strict";
     if (typeof define === 'function' && define.amd) {
@@ -32,10 +27,10 @@
     ];
 
     // Cross-browser bind equivalent that works at least back to IE6
-    function bindMethod(obj, methodName) {
+    function bindMethod(obj, methodName,prefix) {
         var method = obj[methodName];
         if (typeof method.bind === 'function') {
-            return method.bind(obj);
+            return method.bind(obj, `[${methodName.toUpperCase()}]`, prefix);
         } else {
             try {
                 return Function.prototype.bind.call(method, obj);
@@ -63,7 +58,7 @@
 
     // Build the best logging method possible for this env
     // Wherever possible we want to bind, not wrap, to preserve stack traces
-    function realMethod(methodName) {
+    function realMethod(methodName,prefix) {
         if (methodName === 'debug') {
             methodName = 'log';
         }
@@ -73,7 +68,7 @@
         } else if (methodName === 'trace' && isIE) {
             return traceForIE;
         } else if (console[methodName] !== undefined) {
-            return bindMethod(console, methodName);
+            return bindMethod(console, methodName, prefix, methodName);
         } else if (console.log !== undefined) {
             return bindMethod(console, 'log');
         } else {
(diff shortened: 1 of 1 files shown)


pimterry:
Hi @itamayo! Thanks for the PR, but I'm afraid this isn't something I'm interested in adding to loglevel.

Loglevel is intended to remain a very simple reliable library, with advanced functionality added separately via plugins. The code here does all look good, but features like this come with a lot of challenges. Changing the default logging output to add the level to every message would require a breaking version change for starters, and then there's a very wide range of ways that users might want to actually configure specific logging prefixes, so it's very hard to create an API that works for everybody. It's best to handle that separately, instead of baking a million options into the library.

Sorry about that, and I do appreciate you taking the time to build this out regardless! If you're interested, there is an existing plugin with some similar features that you might find useful at https://github.com/kutuluk/loglevel-plugin-prefix. If that doesn't work for your use case, I'd suggest opening a PR there to extend it further, or creating your own plugin from scratch.

If you do publish a plugin then do please let me know, I'd be happy to add it to the readme here so others can easily find & use it.

itamayo:
Hi @pimterry !

I understand and I'm agree with you, but at least the official plugins doesn't allow to trace the log lineNumber what is a problem I'had when I try using this library. In fact I guess the problem could be the plugin System:

https://github.com/kutuluk/loglevel-plugin-prefix/blob/master/lib/loglevel-plugin-prefix.js#L106

Console.X will always run at that line, show trace lineNumber will be wrong....

Anyway thanks for this cool library !

itamayo:
I've check out If there is any possibility of writing a plugin for this scenery, but the case is that the main API should expose these functions to achieve it:

realMethod, bindMethod, replaceLoggingMethods

pimterry:
> Console.X will always run at that line, show trace lineNumber will be wrong....

That's because the specific plugin you linked to calls `originalMethod.apply` there to call the inner logging method inside its own wrapper. It could equally well use `return originalMethod.bind(...)` instead to preserve the stacktraces (for the use cases where bind is sufficient, at least). There are tradeoffs here, but they're not due to limitations in the plugin system AFAIK.

Take a look at the [plugin docs](https://github.com/pimterry/loglevel#writing-plugins) to see how the method wrapper works - you can build the logging methods any way you like. The specific methods you mentioned aren't exposed directly, but I'm pretty sure it's possible to call the built-in method factory function and then wrap the result to do the same thing as this PR.

itamayo:
Thanks for the info, I'll check out deeper ...

itamayo:
Finally, I updated the existing plugin in order to show lineNumberURI adding a small trick. I made PR but I don't know if that plugin is active anymore since there is no activity for more than 3 years. So if it doesn't respond to my PR I'll create a forked plugin.

yracnet:
Hi @itamayo, at final your PR for fixed lineNumber has been accepted for loglevel-plugin-prefix? or Did you publish a new plugin?

itamayo:
I forked a loglevel repo for using muy change since the plugin owner didnt
response the PR, web are using It at production:

https://github.com/itamayo/loglevel

yracnet:
Thanks @itamayo, I will look up others solutions still

yracnet:
Hi @itamayo, when I check the webpack-log I found that this plugin use other library, loglevelnext that is a fork of this repository. https://www.npmjs.com/package/loglevelnext

itamayou:
mmm I didn't know it, thanks

PR #172, opened 2021-11-11
Allow the persistent storage to be cleared

illandril:
This PR adds an optional `{ reset: true }` persist argument value for `setLevel`. This can be used when resetting a log level back to the default state, so that future page loads will fall back to the default log level. This both reduces the clutter in Local Storage and allows people who aren't actively wanting a specific log level to get new default log levels in the (probably unlikely) event the default level is increased or decreased.

In our case, we have a large number of loggers, all of which currently defaults to WARN level. It is not uncommon for someone to want to change a few individual loggers to INFO or DEBUG level to test something or help diagnose an issue, and then reset them back to the default.
This means developers can easily end up with dozens of `loglevel:name` entries in their Local Storage all set to `WARN`, making it harder to find actually useful Local Storage entries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,8 @@ The loglevel API is extremely minimal. All methods are available on the root log
   * As a string, like 'error' (case-insensitive) ← _for a reasonable practical balance_
   * As a numeric index from 0 (trace) to 5 (silent) ← _deliciously terse, and more easily programmable (...although, why?)_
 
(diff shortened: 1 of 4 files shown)


pimterry:
I can see the use case, and the code here looks good at a quick glance, but I'm not totally sure about the design. I'd like to avoid `setLevel()` getting more complicated, and in this specific use case it doesn't seem like you need to change the level - really you just want persistence reset functionality, so they don't need to be linked.

Would adding a `resetLevel()` method instead (resetting the level to default & clearing the persisted level) work for you?

To make that possible, we'd need `setDefaultLevel` to actually store the default level when called (at the moment it's a no-op if there's already a persisted level) but that looks easy to do.

What do you think?

illandril:
See #173

PR #173, opened 2021-11-17
Added resetLevel method

illandril:
This provides a `resetLevel()` method as an alternative solution to the use cases detailed in #172


pimterry:
Yep, I think this looks great, merged. Thanks for the contribution! I'll put out a v1.8.0 release with this included in just a moment.

PR #179, opened 2022-11-02
fix: methodFactory type

grizio:
After investigating the type and the code, I found out that the `methodFactory` type was too permissive.

`methodFactory` is called only for the list of defined methods (from `trace` to `error`).

This commit does not change the javascript logic, only the type definition.

ℹ️ breaking: only in typescript type definition (so only compile-time, not runtime)


pimterry:
Nice improvement, thanks @grizio! Merged, I'll release a patch update shortly.

Issue #123, opened 2018-09-02
No DEMO link in readme


ttodua:
the readme is missing one of the most important thing - it doesnt contain demo.

for example, the BlackBird.js demo - http://demo.jb51.net/js/Blackbird/index.html

see how well it demonstrates the behavior.

pimterry:
It's not the most thrilling library for demos, but I can see how a working live implementation you can play with might be useful. I'd happily accept PRs for this.

hirako2000:
I tried to submit a PR, but I'm having issues actually getting a demo to run as expected.
Here is the value of demos:  Demonstrate basic functionalities. I can't get tracing to work even though I set the level to tracing. A demo can best show how enabling tracing works as I surely do something wrong.

hirako2000:
Do you see the tracing call in the  console tab ? I don't with  latest desktop chrome build for latest osx.

I wouldn't say it's a library bug. Just trying to get close to a demo url we could place on the reademe to keep new users happy.
If it works for you, just merge the codepen. Or provide a better one than the rudimentary statements I placed in there.

bladeski:
@pimterry - This could probably be closed now!

PR #181, opened 2022-12-14
Created a demo page

bladeski:
As per request in #123 this contains a demo of loglevel. It is a static HTML page running some basic JS to showcase some of the features of loglevel, though it does rely on the user checking the console (understandably). I have also updated the README 'Documentation' section to have sub headings for each of the methods. This means that the demo page can link directly to the documentation where appropriate. If this goes through, I'd suggest that using a GitHub action to copy the files to GitHub pages would be the easiest way of keeping the demo with the rest of the code/docs.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -1,0 +1,139 @@
+<!DOCTYPE html>
+<html lang="en-GB">
+    <head>
+        <meta charset="utf-8" />
+        <title>loglevel Demo</title>
+        <meta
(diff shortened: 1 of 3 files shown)


pimterry:
Very nice, thanks, this looks perfect!

I'll merge now, and then I think it should be available via GitHub Pages automatically, I'll just add a link to the README. Great work :+1:

PR #189, opened 2024-01-05
Add `Logger.rebuild()` Method

### LINKED ISSUE #187: Why do I have to configure the log in every typescript file?
A bit of a noob question, but I have to do the following in every `.ts` file that I have. If one file does not do `log.enableAll()` no log is collected for that file.

```typescript
import log from 'loglevel';
import logPrefix from 'loglevel-plugin-prefix';

logPrefix.reg(log);
logPrefix.apply(log, { template: '[%t] %l %n:' });
log.enableAll();
const logger = log.getLogger('DefaultUiAutomationWorker');
```

Why is that?

Mr0grog:
This resolves #187 by adding a new `Logger.rebuild(includeChildren)` method, which takes care of using the `methodFactory` to update all the logging methods according to the current level. If the `includeChildren` argument is set to `true` and the method is called on the root logger, it will recursively rebuild all the child loggers (it is `false` by default, but in some future v2, it should probably be changed to `true`).

The `includeChildren` argument is mainly useful for two cases:

1. When applying a plugin or modifying the logger’s `methodFactory`, it can be used to update previously existing child loggers, so you can divorce logger configuration code from logger declaration code. Ideally, plugin developers will call this from their plugin’s apply/register/whatever code. But a user could call it directly if a plugin developer does not.

2. When changing the level on the root logger and you want that new level to apply to already existing child loggers (that you haven’t called `setLevel()` on and that didn’t have persisted levels). This essentially lets you treat a logger’s “default” level as if it means “whatever my parent/root logger’s current level is.”

So in #187, the example problem code could be updated slightly to make everything work right:

```js
// In `a-random-submodule.js`
import log from 'loglevel'

const logger = log.getLogger('DefaultUiAutomationWorker');
```

```js
// In `index.js`
import aRandomSubmodule from 'a-random-submodule.js';
import log from 'loglevel';
import logPrefix from 'loglevel-plugin-prefix';

function main() {
    logPrefix.reg(log);
    logPrefix.apply(log, { template: '[%t] %l %n:' });
    log.enableAll();
    log.rebuild(true);  // Adding this line is the only change.

    // start rest of program here
}

main();
```

### Other Notes

- In the discussion on #187, we talked about calling this `rebuildLogger()`. I named it `rebuild()` here because I thought that was just as clear and a bit more concise, but I’m happy to change it if you’d prefer.

- I made a small note in the code that `includeChildren` should default to `true` in some future v2 of the library. As discussed in #187, having `setLevel()` affect pre-existing child loggers is probably the right ideal, but might cause subtle breakage, so for now I just made `rebuild()` only affect the current level by default. Flipping the default in the future will also affect the behavior of `setLevel()` and `setDefaultLevel()` since they use `rebuild()` internally. 

    Another option here is to make the default `true` but having `setLevel()` and `setDefaultLevel()` call it with `false` (so `log.rebuild()` affects child loggers by default, but `setLevel()` and `setDefaultLevel()` do not). I think either way would probably be fine. 🤷 

- There are a few subtle side-effects here:

    - The `currentLevel` private variable used to always be a valid level. It no longer is — you should always call `logger.getLevel()` to get the current level. (`defaultLevel` will always be set and valid; `currentLevel` will only be set if `setLevel()` was explicitly called or a persisted level was loaded.) This shouldn’t affect the public interface at all, and I’ve made sure all internal callsites do the right thing.

    - `setLevel()` throws a `TypeError` instead of a `string` for invalid input. I don’t think this should really be an issue, and it arguably improves errors slightly (IMO throwing or rejecting with things that aren’t instances `Error` or a sub-type is problematic). But I can change it back if you prefer.

### DIFF
--- a/lib/loglevel.js
+++ b/lib/loglevel.js
@@ -118,10 +118,30 @@
                enableLoggingWhenConsoleArrives.apply(this, arguments);
     }
 
-    function Logger(name, defaultLevel, factory) {
+    function Logger(name, factory) {
+      // Private instance variables.
       var self = this;
-      var currentLevel;
-      defaultLevel = defaultLevel == null ? "WARN" : defaultLevel;
+      /**
+       * The level inherited from a parent logger (or a global default). We
+       * cache this here rather than delegating to the parent so that it stays
+       * in sync with the actual logging methods that we have installed (the
+       * parent could change levels but we might not have rebuilt the loggers
+       * in this child yet).
+       * @type {number}
+       */
+      var inheritedLevel;
+      /**
+       * The default level for this logger, if any. If set, this overrides
+       * `inheritedLevel`.
+       * @type {number|null}
+       */
+      var defaultLevel;
+      /**
+       * A user-specific level for this logger. If set, this overrides
+       * `defaultLevel`.
+       * @type {number|null}
+       */
+      var userLevel;
 
       var storageKey = "loglevel";
       if (typeof name === "string") {
@@ -185,7 +205,6 @@
           // Use localStorage if available
           try {
               window.localStorage.removeItem(storageKey);
-              return;
           } catch (ignore) {}
 
           // Use session cookie as fallback
@@ -230,31 +249,36 @@
       self.methodFactory = factory || defaultMethodFactory;
 
       self.getLevel = function () {
-          return currentLevel == null ? defaultLevel : currentLevel;
+          if (userLevel != null) {
+            return userLevel;
+          } else if (defaultLevel != null) {
+            return defaultLevel;
+          } else {
+            return inheritedLevel;
+          }
       };
 
       self.setLevel = function (level, persist) {
-          level = normalizeLevel(level);
-          currentLevel = level;
+          userLevel = normalizeLevel(level);
           if (persist !== false) {  // defaults to true
-              persistLevelIfPossible(level);
+              persistLevelIfPossible(userLevel);
           }
-          // NOTE: in v2, this should call rebuild(), which is recursive.
+
+          // NOTE: in v2, this should call rebuild(), which updates children.
           return rebuildLoggingMethods();
       };
 
       self.setDefaultLevel = function (level) {
           defaultLevel = normalizeLevel(level);
-          if (currentLevel == null) {
-              // NOTE: in v2, this should call rebuild(), which is recursive.
-              rebuildLoggingMethods();
+          if (!getPersistedLevel()) {
+              self.setLevel(level, false);
           }
       };
 
       self.resetLevel = function () {
+          userLevel = null;
           clearPersistedLevel();
-          currentLevel = null;
-          self.setDefaultLevel(defaultLevel);
+          rebuildLoggingMethods();
       };
 
       self.enableAll = function(persist) {
@@ -266,24 +290,30 @@
       };
 
       self.rebuild = function () {
+          if (defaultLogger !== self) {
+              inheritedLevel = normalizeLevel(defaultLogger.getLevel());
+          }
           rebuildLoggingMethods();
 
           if (defaultLogger === self) {
-              var level = self.getLevel();
-
               for (var childName in _loggersByName) {
-                _loggersByName[childName].setDefaultLevel(level);
+                _loggersByName[childName].rebuild();
               }
           }
       };
 
-      // Initialize with the right level
-      var initialLevel = getPersistedLevel();
-      if (initialLevel == null) {
-          self.setDefaultLevel(defaultLevel);
-      } else {
-          self.setLevel(initialLevel, false);
+      // Initialize all the internal levels.
+      inheritedLevel = normalizeLevel(
+          defaultLogger ? defaultLogger.getLevel() : "WARN"
+      );
+      if (defaultLevel != null) {
+          defaultLevel = normalizeLevel(defaultLevel);
       }
+      var initialLevel = getPersistedLevel();
+      if (initialLevel != null) {
+          userLevel = normalizeLevel(initialLevel);
+      }
+      rebuildLoggingMethods();
     }
 
     /*
@@ -296,13 +326,15 @@
 
     defaultLogger.getLogger = function getLogger(name) {
         if ((typeof name !== "symbol" && typeof name !== "string") || name === "") {
-          throw new TypeError("You must supply a name when creating a logger.");
+            throw new TypeError("You must supply a name when creating a logger.");
         }
 
         var logger = _loggersByName[name];
         if (!logger) {
-          logger = _loggersByName[name] = new Logger(
-            name, defaultLogger.getLevel(), defaultLogger.methodFactory);
+            logger = _loggersByName[name] = new Logger(
+                name,
+                defaultLogger.methodFactory
+            );
         }
         return logger;
     };
--- a/lib/loglevel.js
+++ b/lib/loglevel.js
@@ -31,6 +31,9 @@
         "error"
     ];
 
+    var _loggersByName = {};
+    var defaultLogger = null;
+
     // Cross-browser bind equivalent that works at least back to IE6
     function bindMethod(obj, methodName) {
         var method = obj[methodName];
@@ -218,10 +221,7 @@
               if (persist !== false) {  // defaults to true
                   persistLevelIfPossible(level);
               }
-              replaceLoggingMethods.call(self, level, name);
-              if (typeof console === undefinedType && level < self.levels.SILENT) {
-                  return "No console available for logging";
-              }
+              return self.rebuild();
           } else {
               throw "log.setLevel() called with invalid level: " + level;
           }
@@ -247,6 +247,27 @@
           self.setLevel(self.levels.SILENT, persist);
       };
 
+      self.rebuild = function (includeChildren) {
+          // NOTE: includeChildren should default to `true` in v2.
+          includeChildren = !!includeChildren;
+
+          replaceLoggingMethods.call(self, currentLevel, name);
+          if (
+              typeof console === undefinedType &&
+              currentLevel < self.levels.SILENT
+          ) {
+              return "No console available for logging";
+          }
+
+          if (includeChildren && defaultLogger === self) {
+              for (var childName in _loggersByName) {
+                var child = _loggersByName[childName];
+                child.setDefaultLevel(currentLevel);
+                child.rebuild(includeChildren);
+              }
+          }
+      };
+
       // Initialize with the right level
       var initialLevel = getPersistedLevel();
       if (initialLevel == null) {
@@ -261,9 +282,8 @@
      *
      */
 
-    var defaultLogger = new Logger();
+    defaultLogger = new Logger();
 
-    var _loggersByName = {};
     defaultLogger.getLogger = function getLogger(name) {
         if ((typeof name !== "symbol" && typeof name !== "string") || name === "") {
           throw new TypeError("You must supply a name when creating a logger.");
--- a/lib/loglevel.js
+++ b/lib/loglevel.js
@@ -86,25 +86,33 @@
 
     // These private functions always need `this` to be set properly
 
-    function replaceLoggingMethods(level, loggerName) {
+    function replaceLoggingMethods() {
         /*jshint validthis:true */
+        var level = this.getLevel();
+
+        // Replace the actual methods.
         for (var i = 0; i < logMethods.length; i++) {
             var methodName = logMethods[i];
             this[methodName] = (i < level) ?
                 noop :
-                this.methodFactory(methodName, level, loggerName);
+                this.methodFactory(methodName, level, this.name);
         }
 
         // Define log.log as an alias for log.debug
         this.log = this.debug;
+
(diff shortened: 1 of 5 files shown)


Mr0grog:
Whoops, forgot to update docs and type definitions. I added two commits to do that.

I’m also noticing that it doesn’t look like Travis is in use any longer, or at least there are no CI checks showing up on this PR. Should we add a GitHub Actions workflow since Travis doesn’t really have a free-for-open-source plan anymore?

lib/loglevel.js
```diff
@@ -245,12 +252,31 @@
           self.setLevel(self.levels.SILENT, persist);
       };
 
+      self.rebuild = function (includeChildren) {
+          // NOTE: includeChildren should default to `true` in v2.
```
pimterry, line 256: Ok, it sounds like this optional argument exists, and this comment, so that this can be called directly for recursive use, and within setLevel for non-recursive use.

I think it would be better to:

* Move the non-recursive logic below into a separate local function, called by both `setLevel` and `rebuild`, so that `setLevel` no longer calls `rebuild` at all.
* Drop the `includeChildren` argument - keep it simple, `rebuild()` always rebuilds children. If there's ever a use case to not do that we can investigate it later but I don't really see when it would be desirable. Since that now only affects the new rebuild API, there's no breaking change.

lib/loglevel.js
```diff
@@ -204,37 +219,29 @@
...
+          return self.rebuild();
       };
 
       self.setDefaultLevel = function (level) {
-          defaultLevel = level;
-          if (!getPersistedLevel()) {
-              self.setLevel(level, false);
+          defaultLevel = normalizeLevel(level);
+          if (currentLevel == null) {
+            self.rebuild();
```
pimterry, line 237: I'm not clear why the defaultlevel logic everywhere needs to change to implement this, can you clarify that? It seems without changing this, rebuild should be able to just call `setDefaultLevel` on children and they would update as expected already.

pimterry (changes requested):

lib/loglevel.js
Mr0grog: Sounds good, will do.

lib/loglevel.js
Mr0grog: Hmmm, I’m not totally sure what you are getting at here, can you clarify a bit? (What do you mean by “the defaulLevel logic everywhere?”)

The problem I encountered going into this is that, under the hood, `setDefaultLevel()` calls `setLevel()`, and `currentLevel` is whatever level the logger is at. BUT that means you can’t tell the difference between a case where the logger is taking its level from its parent vs. a case where a user has explicitly set the logger’s level. There were three solutions that I could think of for that:

1. Have `currentLevel` be the explicitly set level (or null/undefined if not set delegating to the parent’s level) and have `defaultLevel` represent the parent’s level (since we don’t have a reference to the parent). This is what I did because it seemed most straightforward to *me*, logic-wise.

2. Have a new private variable named `hasExplicitLevel` or similar that indicates whether `currentLevel` comes from a user’s call to `setLevel()` (if true) vs. if it came from the parent logger (if false).

3. Have a logger keep a reference to its parent and drop the `defaultLevel` private variable. This has an issue where `getLevel()` might return something that doesn’t match the actual current behavior (if a parent’s level has changed but this logger has not been rebuilt).

(That said, I suspect most of the above is probably clear to you, and I’m missing what you’re getting at.)

Mr0grog:
> I’m also noticing that it doesn’t look like Travis is in use any longer, or at least there are no CI checks showing up on this PR. Should we add a GitHub Actions workflow since Travis doesn’t really have a free-for-open-source plan anymore?

@pimterry Just a quick re-up on the above question. Since #190 landed, I *think* this should be pretty easy to do.

lib/loglevel.js
Mr0grog: OK, this should be fixed in 45396397b87dacc20a62bc9bf303a70c61854be1. I also rebased this PR since #190 landed, which required some very minor test changes.

One quick question this brought up for me: should the new private method here just get merged together with `replaceLoggingMethods()`? We now have:

```js
    // Expects to be called with apply/call to set `this` to the logger.
    function replaceLoggingMethods(level, loggerName) {
        /*jshint validthis:true */
        for (var i = 0; i < logMethods.length; i++) {
            var methodName = logMethods[i];
            this[methodName] = (i < level) ?
                noop :
                this.methodFactory(methodName, level, loggerName);
        }

        // Define log.log as an alias for log.debug
        this.log = this.debug;
    }

      // Private to each logger; uses `self` instead of `this`
      function rebuildLoggingMethods() {
          var level = self.getLevel();

          replaceLoggingMethods.call(self, level, name);
          if (typeof console === undefinedType && level < self.levels.SILENT) {
              return "No console available for logging";
          }
      }
```

It doesn’t seem obvious to me that these shouldn’t just be one function; I don’t feel like there’s much responsibility difference between them other than `rebuildLoggingMethods()` checking whether `console` is present, but I’m not sure it would be bad if `replaceLoggingMethods` did that). The only other callsite for `replaceLoggingMethods()` is in `enableLoggingWhenConsoleArrives()`, and I think it *might* be better/simpler for that to be calling `rebuildLoggingMethods()` anyway.

pimterry: I don't have a strong preference here, whichever you prefer is fine by me.

lib/loglevel.js
pimterry: I think this should also applies to `setDefaultLevel` though.

Reading the code, I think the current model implemented there is that 'something persisted' means that the value was overridden (and so shouldn't be changed) and anything else means that the value is a default value which should change when the default level changes.

We shouldn't change that existing logic for default levels (whether or not it's a good design, it's a breaking change that would unexpectedly change what's logged by people's applications).

I think the same logic should simply apply for the new rebuild API: if the pre-existing setDefaultLevel logic would modify the level, then rebuilding should modify the level.

In a future version we could re-evaulate this, and in that case I think there's a good argument for explicitly tracking whether the level was set manually or not, but for now I think we have to stick with the existing behaviour.

pimterry:
> > I’m also noticing that it doesn’t look like Travis is in use any longer, or at least there are no CI checks showing up on this PR. Should we add a GitHub Actions workflow since Travis doesn’t really have a free-for-open-source plan anymore?

> @pimterry Just a quick re-up on the above question. Since https://github.com/pimterry/loglevel/pull/190 landed, I think this should be pretty easy to do (and I am happy to do the work; just let me know).

Sorry I missed that - yes, Travis is indeed dead, I'd be extremely happy to move to GitHub Actions if you want to set that up.

lib/loglevel.js
Mr0grog: > Reading the code, I think the current model implemented there is that 'something persisted' means that the value was overridden (and so shouldn't be changed) and anything else means that the value is a default value which should change when the default level changes.

I guess I’d argue that leaning on persistence is already problematic, since it doesn’t work in environments that have no persistence mechanism (e.g. Node, Bun, an `iframe` with a lot of sandboxing options turned on, etc.). It means this happens, which sure feels like a bug to me:

```js
var child = log.getLogger("test");
child.getLevel();  // 3 (WARN)

child.setLevel("TRACE");
child.getLevel();  // 0 (TRACE), good!

child.setDefaultLevel("DEBUG");
child.getLevel();  // 1 (DEBUG), seems bad!
```

Or more saliently, once there’s a `rebuild()` method:

```js
log.getLevel();  // 3 (WARN)
var child = log.getLogger("test");
child.getLevel();  // 3 (WARN), it inherited, good.

child.setLevel("TRACE");
child.getLevel();  // 0 (TRACE), good!

log.rebuild();
child.getLevel();  // 3 (WARN), seems bad that rebuilding changes the
                    // child's level after you've explicitly set it.
```

I suppose this is, very subtly, breaking, the question is whether the current behavior is a bug (my view, and hence why it is OK to “fix” as a byproduct of this) or is just what it is. I think most people’s mental model here is what I’m doing in this change and so wouldn’t be surprised by it.

Mr0grog: There is also a test that models the above scenario: https://github.com/pimterry/loglevel/blob/45396397b87dacc20a62bc9bf303a70c61854be1/test/multiple-logger-test.js#L175-L182

I feel like it is clear to me that that test should always succeed (and therefore needs this change, or one of the similar things I enumerated above).

Mr0grog: Ugh, looked back at the discussion when I originally added `setDefaultLevel()`, and I am definitely wrong. “Default’ was probably a confusing name choice (my fault!) and I was over-focused on the browser context at the time. Anyway, I had lost track of the fact that `setDefaultLevel()` was originally intended to be a way to say “set the level unless you’ve loaded saved configuration info from somewhere,” *not* “set the level if one hasn’t been explicitly set otherwise.” The idea was that `setDefaultLevel()` was for built-in config in your app, and `setLevel()` was only for runtime *re*configuration by users (e.g. a dev setting it in their console, some preference a user might adjust, a CLI flag).

Obviously I need to re-do this. `setDefaultLevel()` might not have been well designed (my fault!), but it is working as designed and is far too well baked to change, as you noted.

> I think the same logic should simply apply for the new rebuild API: if the pre-existing setDefaultLevel logic would modify the level, then rebuilding should modify the level.

I don’t think this is quite right, though. With the right definition for `setDefaultLevel()` (above), it seems like it obviously qualifies as explicitly setting the logger’s level just as much as `setLevel()` does. Using either one means the logger’s level shouldn’t be influenced by its parent anymore.

That is:

```js
log.setLevel("WARN");

var child1 = log.getLogger("child1");
child1.setLevel("ERROR");
child1.getLevel();  // ERROR

var child2 = log.getLogger("child2");
child2.setDefaultLevel("ERROR");
child2.getLevel();  // Let's assume ERROR and something was not previously saved

var child3 = log.getLogger("child3");
child3.getLevel():  // Let's assume it WARN because nothing was previously saved

log.methodFactory = function() { ... }
log.setLevel("TRACE");
log.rebuild();

// So the levels should now be:
//   log    = TRACE (we just set it)
//   child1 = ERROR (was set with setLevel)
//   child2 = ERROR (was set with setDefaultLevel)
//   child3 = TRACE (was never set, so inherits the change from the parent)
```

Let me know if you think the above doesn’t make sense, but it seems right to me. I think the disconnect here is over what level `child2` should be at.

This also makes it a bit ambiguous what `resetLevel()` should do. As coded today, calling it on the children would change `child1` to TRACE and `child2` would remain at `ERROR`. I think that’s probably fine — it resets from the *persisted* level to the logger’s expected level — but I can imagine people expecting it clears any level info on the logger in favor of inheriting from the parent instead.

Mr0grog: OK, this should be solved. I kept getting myself twisted up in the logic with the boolean, so I wound up refactoring this as 3 private variables. LMK know if you find this too confusing and I will refactor some more, but it felt much easier to comprehend for me:

- `inheritedLevel` is a cache of parent logger's level (or `WARN` for the root logger). This is always set to something valid.
- `defaultLevel` is the default level that has been set for this logger. It may be null (unset).
- `userLevel` is the level that has been set for this logger via persistence or via `setLevel()`. It may be null (unset).

The current/effective level of the logger is `userLevel OR defaultLevel OR inheritedLevel`.

This seemed like a more straightforward model of what’s going on, and the conceptual order of precedence is relatively obvious. There's no boolean or other state that is modifying how the levels behave based on what methods have been called in what contexts (`resetLevel()` was causing me subtle problems, although after the above renaming, I am pretty sure I could switch back to having the boolean and get it right this time).

---

The other thing worth noting here is that I opted to have child loggers reach up to their parent to get the parent’s level, where I previously had the parent updating the children’s levels. Based on the discussion here, we would have had to add a special public method for the parent to call, which seemed bad. So I flipped it around.

(Side note: since we only have one level of children right now, the children can just grab the parent level from `defaultLogger`. In some hypothetical future with a deep hierarchy of loggers, each logger will need to hold a reference to its parent.)

Mr0grog:
Alright, this should be ready for another review (and maybe to land). See my replies to the previous review comments for an update on the architecture here: https://github.com/pimterry/loglevel/pull/189#discussion_r1462567157

I added more comments, which pumps up the un-minified size a bit:

| File              |  Before |    After | Change |
| ----------------- | ------: | -------: | -----: |
| `loglevel.js`     | 9.41 KB | 11.04 KB | +17.3% |
| `loglevel.min.js` | 3.08 KB |  3.26 KB |  +5.9% |

pimterry (approved):

pimterry:
Thanks, I agree that 3 variable model looks cleaner, and I'm happy this pulls together the rebuild logic without changing the existing code too significantly.

Thanks for all your hard work to make this and the other PRs happen. It's really tidied up the project and should be a great fix to help others who might be affected by #187 and similar confusion. Great stuff :+1:. I'll do some last testing & checks locally and then publish a new release with these changes included soon.

Mr0grog:
Awesome! 🙌 

I’ll finish up with a commit that updates the contributing instructions as mentioned in https://github.com/pimterry/loglevel/issues/191#issuecomment-1894653099.

PR #190, opened 2024-01-09
Update dev deps to make tests work on MacOS and resolve Critical arbitrary code execution/command injection vulnerabilities

Mr0grog:
While working on #189, I discovered I couldn’t run tests locally on any of my Macs (both x86/Intel and ARM/Apple machines). Installing the dev deps also fired off a laundry list of warnings about security vulnerabilities, including several critically rated ones about arbitrary code execution or command injection.

Generally speaking, Loglevel doesn’t do much at build/test time and doesn’t read arbitrary input that could compromise things, but that doesn’t mean some transitive dependencies don’t read some config from surprising places (many tests also run in a version of PhantomJS that has and arbitrary file read vulnerability, and any network traffic that might happen to occur in Phantom could potentially be a route for compromise). Anyway, I don’t think the risks are super high here, but they aren’t non-existent, and are probably important to fix. Plus, I’d like to be able to develop locally instead of on a remote linux machine, which I had to do here. 🙂

So! I generally made the minimum updates possible here that fix critical arbitrary execution or command injection vulnerabilities (there are still other critical vulns left here, but they are IMO not a big risk given the dev-time-only context we are in).

I also only upgraded things as far as they could go without requiring a Node.js version >0.10.0. That said, this is still newer than v0.6.0, which is what this package aims to be compatible with, meaning we cannot run tests on all versions of Node.js that we want to support. But:
- The existing deps *before* this PR already all require either v0.8.0 or v0.10.0 (and if you count the Typescript stuff, v4.2.0), so this doesn’t really change Node.js requirements in practice.
- I could not successfully install the existing deps *before* this PR on Node.js <4 and could not get the tests to run on Node.js < v6 (I tried 0.6.x, 0.8.x, 0.10.x, 0.12.x, 4.x, 6.x, 8.x, and 10.x). For this PR, I installed and tested on Node.js v6 and v8. *I could probably have upgraded some deps farther if I only worried about Node.js 6+ compatibility, but did not do that here — please let me know if I should.*
- Some transitive deps declare requirements for even higher versions of Node.js (e.g. v10). Things in the test `npm test` and build (`npm run dist`) toolchain seem to work fine, so I ignored these warnings. (And the warnings predate this PR anyway.)

This updates Uglify, which results in differences to `dist/loglevel.min.js`, but everything seems to still work fine, so I think this is safe.

There are also a bunch of test updates because the only versions of grunt-contrib-jasmine that solved issues upgrade from Jasmine v1 to v2:
- The way you declare and add custom matchers is different.
- Tests with no expectations are treated as failing. I added `expect().nothing()` calls in these cases. That expectation was added to Jasmine specifically to use in these cases.
- Async support is much different, so I had to rewrite the loglevel reloading helper.
- Unfortunately, the Node.js tests still use Jasmine v1, which could make for some confusion when working on tests. The plugin for it, grunt-jasmine-node, is no longer maintained, and the underlying jasmine-node package is only maintained for security updates, and does not intend to support newer Jasmine versions. They recommend just using the Jasmine CLI. I can add that here or in a follow-on PR if you want (actually I think we could call it through Grunt, but either way, we have to do it ourselves without any plugin magic).

Finally, the biggest change is that I had to vendor `grunt-template-jasmine-requirejs` and alter some of the code in it. The package is no longer maintained, and no published versions are compatible with Grunt 1.x. Every change I made to it is [marked with `// LOGLEVEL-FORK: <description>` comments](https://github.com/Mr0grog/loglevel/blob/5f3764e7b1523c8052d3e30f3bb5f63933d6cfb0/vendor/grunt-template-jasmine-requirejs/README.md?plain=1#L1-L11). For example: https://github.com/Mr0grog/loglevel/blob/5f3764e7b1523c8052d3e30f3bb5f63933d6cfb0/vendor/grunt-template-jasmine-requirejs/src/template-jasmine-requirejs.js#L193-L195

I chose to include the entire package source when doing this. Alternatively, we could:
1. Just include the relevant parts (no tests, README, `package.json`, etc.), or
2. Write our own more minimal module for this, since we don’t need to cover all the generic cases the plugin does.
3. We could also switch to [@radum/grunt-template-jasmine-requirejs](https://www.npmjs.com/package/@radum/grunt-template-jasmine-requirejs), but I’m not sure whether the long-term support story will be any better (it has 1 download/wk and hasn’t been touched in 6 years) and it requires Node.js 4+. I think it is probably wiser/more future-proof to maintain our own support here.


Mr0grog:
Also! This does not work in Node.js 12+ (that is, it works in Node.js 6.0.0–10.x inclusive) There appear to be some issues with newer Node.js versions in either jasmine-node or Grunt somewhere that need more investigation. I figured it was better to get this incremental improvement in first.

Mr0grog:
Al so just realized I did not clarify this issue and its fix:

> I couldn’t run tests locally on any of my Macs (both x86/Intel and ARM/Apple machines).

The core problem here was an old version of PhantomJS. The updates in this PR cause us to install and use PhantomJS v2.x, which works on current versions of MacOS.

(Worth noting: the PhantomJS project is very dead, and newer versions of the tools that were using it have switched to Pupppeteer/headless Chrome. Those updates require newer versions of Node.js, though, so I did not update to them here.)

PR #192, opened 2024-01-11
Fix cookie persistence, also fix some tests that are broken in modern browsers

Mr0grog:
This fixes a bug where any levels that were saved in cookies were not loaded correctly. It also fixes a bunch of tests that were broken only in live/modern browsers and not in Node.js or PhantomJS (what you get when you run `npm test`). To see these failures, you would have had to start a server with `grunt integration-test` and then opened the Jasmine test runner from the server that starts in a real browser.

The cookie issue is that we were reading the value starting from the beginning of the cookie’s key-pair, so instead of reading a value like `"INFO"`, we’d read `loglevel="INFO"`, which is obviously not a valid level.

This also fixes a number of other test failures when running tests in an actual browser, but all of them were problems with the tests rather than actual bugs in the library like the above cookie issue. Fixing the cookie issue also revealed some *other* bugs in the test, so this was a bit of an iterative process.

Another thing I did with the tests here was change the `describeIf`/`itIf` helpers to mark tests that don’t get run as skipped rather than hiding them from test output entirely, so it’s at least a little more obvious that some things are not getting tested.

### DIFF
--- a/lib/loglevel.js
+++ b/lib/loglevel.js
@@ -158,10 +158,12 @@
           if (typeof storedLevel === undefinedType) {
               try {
                   var cookie = window.document.cookie;
-                  var location = cookie.indexOf(
-                      encodeURIComponent(storageKey) + "=");
+                  var cookieName = encodeURIComponent(storageKey);
+                  var location = cookie.indexOf(cookieName + "=");
                   if (location !== -1) {
-                      storedLevel = /^([^;]+)/.exec(cookie.slice(location))[1];
+                      storedLevel = /^([^;]+)/.exec(
+                          cookie.slice(location + cookieName.length + 1)
+                      )[1];
                   }
               } catch (ignore) {}
           }
--- a/test/default-level-test.js
+++ b/test/default-level-test.js
@@ -23,6 +23,10 @@ define(['test/test-helpers'], function(testHelpers) {
         });
 
         describe("If no level is saved", function() {
+            beforeEach(function () {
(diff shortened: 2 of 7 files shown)


lib/loglevel.js
```diff
@@ -158,10 +158,12 @@
...
                   var cookie = window.document.cookie;
-                  var location = cookie.indexOf(
-                      encodeURIComponent(storageKey) + "=");
+                  var cookieName = encodeURIComponent(storageKey);
+                  var location = cookie.indexOf(cookieName + "=");
                   if (location !== -1) {
-                      storedLevel = /^([^;]+)/.exec(cookie.slice(location))[1];
+                      storedLevel = /^([^;]+)/.exec(
+                          cookie.slice(location + cookieName.length + 1)
+                      )[1];
```
Mr0grog, line 166: This is the actual bugfix. The rest is just cleaning up bad test code.

pimterry (approved):

lib/loglevel.js
Mr0grog: > Yep, this makes sense. Any idea why this worked in phantomjs though?

It didn't! One of the reasons I switched to showing the tests as skipped was to verify for myself that *none of the cookie tests run in PhantomJS*. It turns out [`document.cookie` is just always empty and cannot be set](https://github.com/ariya/phantomjs/issues/11222), which causes all the cookie-related and persistence-related tests to be skipped.

Issue #191, opened 2024-01-10
Cleanup/modernization of developer test/build tooling


Mr0grog:
This is a follow-on from #190, where I did some basic work to move the dev tooling forward just a bit and resolve some [slight] potential supply-chain style security risks. There were a bunch of open questions about what is worth doing or should be done for dev tooling going forward, and it seems better to have an open issue for this rather than discussion in a completed PR.

So, continuing the discussion from https://github.com/pimterry/loglevel/pull/190#issuecomment-1885261203 …

> while [more updates] are very helpful, I don't think it's worth trying to incrementally update this further for modern node etc.

FWIW, I would like to try and update a little further here (that is, I am happy to do the work if you don’t mind reviewing), since it is still tough to get Node.js 10 set up on a Mac with an Apple/ARM CPU (i.e. my main computer 😉). I also think enabling tests on current versions of browsers and Node is good because as much as JS engine developers don’t want to break the web/break existing code, sometimes it happens.

Also, I just noticed that the test suite is currently passing in PhantomJS, but *not in modern browsers* (I had not tried viewing Jasmine tests manually via the `integration-test` task before). So that’s probably good evidence this needs some more updates. 😬 

> It's going to be a significant amount of work, and it would probably be better spent just replacing the test setup entirely, to rewrite it completely with modern tools. Or, more realistically, as loglevel is mostly in 'done' maintenance don't-change-too-much mode

Yeah, agree that loglevel is mostly in maintenance, and working *too* hard on this, or making changes that are too big are probably not worthwhile. What I think is worth focusing on here is making the DX tooling a little more resilient to to ravages of time:

- Update to the latest Grunt. Grunt is in a similar “pretty stable and done, but still well maintained for security/bugs” status as loglevel.
- Or drop Grunt if that turns out to be easier than maintaining a Grunt-based setup [because of outdated plugins]. I think I’ll discover which is easier pretty quickly when doing the work.
- We probably don’t want really fancy new tooling, because that tooling is more likely to change a lot need a lot of updates by the next time a feature or fix needs to reviewed here.
- Basically: minimal tooling with minimal tooling-caused churn.

Having a test suite working on modern Node & browsers also makes every change and bugfix PR submitted here more safe. That’s only more valuable for something in maintenance mode.

> In any world where somebody has the time & interest to build out loglevel v2 and take this forward with more substantial changes, I think the first step would be creating a new build env for modern node & tools, starting from a mostly clean slate.

[...]

PR #193, opened 2024-01-12
Add basic CI workflow on GitHub Actions

Mr0grog:
This adds a "CI" workflow on GitHub Actions to run all the various tests on every PR. I broke out the major types of tests and test environments into different jobs so you can see which things are passing and failing in the PR checks more clearly:
- Linting (JSHint)
- Building
- Browser-based (currently PhantomJS) tests
- Node.js-based tests
- TypeScript type tests

~This might need GitHub actions to be turned on in the settings for this repo (and also turned on for forks).~ *(Update: nevermind, looks like this is already done!)*

Based on discussion in #191.


pimterry:
Nice work, this looks great, thanks!

PR #194, opened 2024-01-12
Shed some vestigial dev tooling

Mr0grog:
~⚠️ **This is based on #193; that PR should be merged first.** ⚠️~

Based on the discussion in #191, this removes a bunch of dev-time tooling that we no longer actually need or use. It also happens to resolve 31 of our vulnerabilities from dependencies (based on the results of `npm audit`), including 4 that were rated “critical” (there are 21 total left, 2 of which are “critical”).

Removed:
- QUnit (this was only used to verify that we don’t cause problems when used by a package that is tested with QUnit; I replaced it with a simple test of `noConflict()`, which is what this is really about)
- Saucelabs
- Coveralls.io (if we want to start using Coveralls.io again, we should probably use [their official GitHub Action](https://github.com/marketplace/actions/coveralls-github-action) instead)
- Vendored `json2.js` package (it never seems to have actually been used)
- Removes a redundant step from the `test-browser` task in Grunt.

There’s still a little more easy-ish cleanup/updates to do based on #191, and I’ll work on that after this PR lands.


pimterry:
Can you rebase this on top of main? That should drop the duplicated CI commits that are still shown here.

Mr0grog:
Done!

PR #195, opened 2024-01-17
Get tests and builds working in current Node.js versions

Mr0grog:
This gets tests and builds working in current versions of Node.js, from v6 through v20 (current active release), and in MacOS on Apple Silicon. **The main goal here is to make it possible to run tests and builds on most types of computers without lots of complex setup, so it’s not too hard for contributors to test and validate their work.**

⚠️ There are some messy bits here, and you might want to see some changes here, or might feel any solutions here are too complex and this PR should not be merged at all.

The main changes here are:

- **Upgrade _grunt-contrib-jasmine_ to v4.0.0 and _grunt-template-jasmine-istanbul_ to v0.6.0 (via git, not NPM)**. Unfortunately, _grunt-template-jasmine-istanbul_ now points to [the git commit for v0.6.0](https://github.com/maenu/grunt-template-jasmine-istanbul/commit/caedc636b9f9c11217a32d586717e3153edf9ba3) instead of NPM, because this version was never published on NPM. I believe that is because of some potential issues with tests in [the final change](https://github.com/maenu/grunt-template-jasmine-istanbul/pull/64#issuecomment-547359194) (it looks like nobody ever verified whether *actually* are issues — the author supposes they might be resolved when another, parallel PR merges, and that PR did in fact get merged).

    Some options on making this situation better:
    1. Drop _grunt-template-jasmine-istanbul_ entirely. This means we lose test coverage data, but we aren’t currently using that data for anything anyway. (I think, given the current status of things, this is the best option. I didn’t do it yet in this PR to keep things clear and minimal.)
    2. Vendor this version of _grunt-template-jasmine-istanbul_ so we aren’t relying on a git repo or commit that could disappear.
    3. Downgrade to _grunt-contrib-jasmine_ v1.2 + _grunt-template-jasmine-istanbul_ v0.4.0. This is less ideal because this version of _grunt-contrib-jasmine_ still uses PhantomJS, which has been dead for 6 years. Eventually it will probably start having problems in newer systems. This version also has a critical command injection vulnerabiltiy, and would reverse some of the work I went through to clean those out in #190. You can see an implementation of this in the first commit of this PR (8a4a74406beafad4523bc650fc1bbbf1e2f193e8).

- **Remove _grunt-jasmine-node_ in favor of using _jasmine_ directly.** The new, inline Grunt task for calling Jasmine directly is extremely simple, and I think it poses low maintenance overhead. It also removes some security issues and allows us to use the same version of Jasmine in both Browser-based tests and Node.js-based tests.

    An alternative here is to use _grunt-jasmine-node_ v0.3.1, but it requires a ridiculous hack because it turns out to be totally broken with respect to loading configuration options. More info in this commit that tries using it: 94acb4ec900aa4ae3ea85485bacb5e88c2a886f4. All versions of _grunt-jasmine-node_ that work in modern Node.js versions have this problem. I would definitely recommend against this.

This completes the first two items from https://github.com/pimterry/loglevel/issues/191#issuecomment-1894653099.


.editorconfig
```diff
@@ -22,3 +22,6 @@ indent_size = 2
 
 [*.{yaml,yml}]
 indent_size = 2
+
+[vendor/grunt-template-jasmine-requirejs/**/*]
+indent_size = 2
```
Mr0grog, line 27: I probably should have done this when I vendored *grunt-template-jasmine-requirejs*. It was helpful when I had to make changes. 🤷

.github/workflows/ci.yml
```diff
@@ -18,12 +18,12 @@ jobs:
...
         uses: actions/setup-node@v4
         with:
-          node-version: 10
+          node-version: 20
           cache: npm
           cache-dependency-path: 'package.json'
 
       - name: Install dependencies
-        run: npm install
+        run: npm install --legacy-peer-deps
```
Mr0grog, line 26: The `--legacy-peer-deps` option is required for *grunt-template-jasmine-istanbul* v0.6.0, since it depends on *grunt-contrib-jasmine* v2.x and we are now on v4.x. It’s less than ideal, but is only needed in order to install dev dependencies; a consumer of loglevel should not need it when running `npm install loglevel` in their package.

As noted in the PR description, I actually think the best solution if we want to remove this is to drop *grunt-template-jasmine-istanbul* altogether.

Mr0grog:
Realized today that in 72182e5e5deb7a8540c7e02f168ee4cbc4669d62 when I updated the version of *grunt-contrib-jasmine* to v4, it also upgrade Jasmine to v5.1 for browser tests. I figured I should upgrade the Node.js tests to match, which was pretty simple, but then realized those require at least Node 18+, which seemed like a bit too big of a jump, so I reversed that. This is easy to put back in with a single `git revert bb00d4193688255c39544cabf493e4bad461651e`, though.

Along the way, I also noticed that cookie support checking in the tests is not actually working correctly, resulting in some tests getting skipped when they shouldn't be (this was fine in PhantomJS since it doesn’t properly support cookies, but different in headless Chrome or in live tests with `grunt integration-test`). I fixed that check in 160fc4cce49e9607148d23fc6e5c9b17099c0553.

pimterry:
> Some options on making this situation better:
> * Drop grunt-template-jasmine-istanbul entirely.

Yeah, I'm fine with that. It seems like the current approach adds complexity (requiring legacy peer deps options, using unofficial releases, depending on a now-unmaintained dead project) and as we discussed in #191 we're not actually using the coverage data now.

If it was easy to keep working that'd be fine, but this seems more hassle than its worth, so I'm happy to cut it entirely and streamline.

.github/workflows/ci.yml
pimterry: This sounds like something that would make contribution difficult though, it would be nice if this worked out of the box so that contributors can just run `npm install` without problems.

You can run `npm config --location=project set legacy-peer-deps=true` to fix this - that'll create a `.npmrc` that will enable this setting by default for any npm command run in this repo, and then you can drop that option from here.

(But as noted in my other comment - I think we should just drop this instead, so you don't actually need to do that unless there's some other legacy peer deps issue elsewhere too)

Mr0grog: Ah, I had not thought about adding a `.npmrc` file for this, good catch! But also, yeah, agree on the downsides anyway.

Mr0grog:
Makes sense. That should be done, and this PR should be ready to go.

Issue #196, opened 2024-01-25
Cannot set property trace of #<Object> which has only a getter


designamx:
starting today we are getting this error
```
[ERROR] 14:32:05 TypeError: Cannot set property trace of #<Object> which has only a getter
TypeError: Cannot set property trace of #<Object> which has only a getter
api.apt.local     |     at Object.replaceLoggingMethods (/usr/src/node_modules/loglevel/lib/loglevel.js:96:30)
api.apt.local     |     at Object.Logger.self.setLevel (/usr/src/node_modules/loglevel/lib/loglevel.js:267:40)
api.apt.local     |     at Object.<anonymous> (/usr/src/api/src/config/logger.ts:6:8)
api.apt.local     |     at Module._compile (node:internal/modules/cjs/loader:1241:14)
api.apt.local     |     at Module._compile (/usr/src/node_modules/source-map-support/source-map-support.js:568:25)
api.apt.local     |     at Module.m._compile (/tmp/ts-node-dev-hook-024130929438604287.js:69:33)
api.apt.local     |     at Module._extensions..js (node:internal/modules/cjs/loader:1295:10)
api.apt.local     |     at require.extensions..jsx.require.extensions..js (/tmp/ts-node-dev-hook-024130929438604287.js:114:20)
api.apt.local     |     at require.extensions.<computed> (/tmp/ts-node-dev-hook-024130929438604287.js:71:20)
api.apt.local     |     at Object.nodeDevHook [as .ts] (/usr/src/node_modules/ts-node-dev/lib/hook.js:63:13)
```

Node: `node:20.6.1-alpine3.17` also tested `node:20.10.0-alpine3.17`

itsjbecks:
Also been having this issue on `Node.js v20.11.0` and `v18.14.2` (very recently)

pimterry:
This must be related to the v1.9.0 version just released - you can downgrade to v1.8.1 to avoid this.

Can you provide a standalone example to reproduce the issue though? I'm not sure exactly what would cause this.

designamx:
@pimterry  thanks, that was it

pimterry:
@designamx this is clearly still an issue in the package, but glad to know the downgrade resolves this for you.

Can you please share an example to help reproduce this, or any more information you can about how you're using loglevel? Otherwise this isn't going to be resolved in future releases of the package and so you'll just hit the same issue again later on.

designamx:
sure, I can do that

designamx:
[...]

PR #197, opened 2024-01-26
Run TypeScript tests in ts-node

Mr0grog:
We previously ran the `test/type-test.ts` test through the TypeScript compiler to catch type issues, but there are more subtle problems that can happen at runtime with tools like `ts-node` (and maybe things like Deno, too — I haven't tested that). This actually *runs* the test file in ts-node to make sure it doesn't throw at runtime, which would have caught issues like #196 ahead of time.

~⚠️ This isn’t ready to merge. I added a commit that *should* cause the test to fail (partly reverting the fix for #196). If it fails, that I’ll take that back out and it should be merge-able.~

### DIFF
--- a/lib/loglevel.js
+++ b/lib/loglevel.js
@@ -264,7 +264,7 @@
           }
 
           // NOTE: in v2, this should call rebuild(), which updates children.
-          return replaceLoggingMethods.call(this);
+          return replaceLoggingMethods.call(self);
       };
 
       self.setDefaultLevel = function (level) {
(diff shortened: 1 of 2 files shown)


Mr0grog:
Ah, beautiful failure! This prints out some pretty good info in the GitHub Actions logs:

Screenshot of logs from CI failure

Mr0grog:
✅ Should be safe to ship now!

pimterry:
Good idea, thanks!

Issue #119, opened 2018-01-31
Loglevel 2.0


kutuluk:
Loglevel is good. It's time to make it fresh. More clearly. More modern.

- Remove support for legacy browsers.
- Remove support for persistance levels.
- Remove noConflict()
- Сustomizable levels
- Flat flow of plugins (only one extra line in the stacktrace for any number of plugins)
- `log.getLogger('child')` => `log('child')`
- getter/setter log.level
- maybe something else...

**Implementation loglevel.js**
```javascript
const noop = () => {};
const loggers = {};
const configs = {};
const chain = [];
let levels = ['trace', 'debug', 'info', 'warn', 'error', 'silent'];
let levelsInfo = levels.slice();

// Build the best logging method possible for this env
// Wherever possible we want to bind, not wrap, to preserve stack traces
function defaultFactory(methodValue) {
  /* eslint-disable no-console */
  if (typeof console === 'undefined') return noop;

  const methodName = console[levels[methodValue]] ? levels[methodValue] : 'log';
  if (console[methodName]) return console[methodName].bind(console);

  return noop;
  /* eslint-enable no-console */
}

function pluginsFactory(methodValue, logger) {
  const methods = [];

  chain.forEach((plugin) => {
    const rootConfig = configs[plugin.name][''];
    const loggerConfig = configs[plugin.name][logger.title];
    if (rootConfig || loggerConfig) {
      methods.push(
        plugin.factory(
          methodValue,
          logger,
          Object.assign({}, plugin.defaults, rootConfig, loggerConfig),
        ),
      );
    }
  });

  const native = defaultFactory(methodValue);

  return (...args) => {
    for (let i = 0; i < methods.length; i++) {
      methods[i](args);
    }
    native(...args);
  };
}

let factory = defaultFactory;

function rebuildMethods(logger) {
  for (let i = 0; i < levels.length - 1; i++) {
    logger[levels[i]] = i < logger.level ? noop : factory(i, logger);
  }
}

function removeMethods(logger) {
  for (let i = 0; i < levels.length - 1; i++) {
    delete logger[levels[i]];
  }
}

function Logger(logName, logLevel) {
  const logger = this || {};

  const defineProperty = Object.defineProperty;

  defineProperty(logger, 'title', {
    get() {
      return logName;
    },
  });

  defineProperty(logger, 'level', {
    get() {
      return logLevel;
    },
    set(lvl) {
      let newLevel = lvl;

      if (typeof newLevel === 'string') {
[...]

PR #203, opened 2024-12-19
v2: First steps, drop old IE support, fix `debug`

### LINKED ISSUE #137: log.debug should map to console debug in Chrome
Although the [documentation](https://developers.google.com/web/tools/chrome-devtools/console/api) for Chrome console specifies that console.debug has a log level of "info", it actually has log level of "verbose". 

With the current loglevel implementation, debug messages are mapped to console.info. This behavior doesn't allow filtering messages in the dev console which are actually "verbose", and hence causing console.debug and log.debug to have different behaviors. 

I suggest that log.debug would map to console debug.

Mr0grog:
This is a very first step towards v2, based on discussion in #119. It just does two main things:

1. Drop support for IE < 11 — lazy console, special `trace` handling, etc.
2. Make `debug` call through to `console.debug` instead of `console.log` (fixes #137).

This does _not_:
- Change tooling or switch to ESM. Those should probably be the next PR.
- Make any changes to persistence. I’m also leaving that to a separate PR.

⚠️ Unfortunately, there is one debatable bit here! We don’t currently have a level reserved for `log`, so I’ve left it as an alias for `debug` (it used to be the other way around). Should we add one? (I think yes.) Where does it sit in terms of levels? Precedent varies:
- The same as `info` (Chrome)
- Between `debug` and `info` (closest to Loglevel’s v1.x behavior)
- Between `info` and `warn` (Safari & Firefox)

I’d vote for “between `info` and `warn`” to match browsers as well as possible.

Other options could include:
- Dropping it since it isn’t exactly a “level” (i.e. not having a `logger.log()` method at all)
- Making it a way to call with a dynamically chosen level (rather than a pass-through to console), i.e. `logger.log(level, ...messages)`)
- Making it have the same *level* as another method but still call `console.log` under the hood (right now it is the same level as `debug`, and also _calls_ `console.debug`). If we did this, maybe best to make it match the `info` level (Since that is how Chrome treats it)?

*(Update: re-formatted all the stuff about `log()` to make it more clear.)*


Ryuno-Ki:
Can you switch the target branch?
The README is claiming that this is going to be an in-development branch, but after merge, this is part of `main` which will therefore no longer be „stable”.

Mr0grog:
Whoops, thought I'd done that! Thanks for catching.

Mr0grog:
> Should we add [a `log` level]? (I think yes.) Where does it sit in terms of levels?

Looking back at this at the end of the day, I think I’ve changed my mind. The README has long made clear that the `.log()` method is *just* there to ease migration from `console` and is not really a “level”:

> `log.log(msg)` is also available, as an alias for `log.debug(msg)`, to improve compatibility with `console`, and make migration easier.

So actually the way this PR is now is probably fine.

The only change I might still suggest is making `log` an alias for `info` instead of `debug` in order to work more like browser dev tools:
1. That is how Chrome treats it and
2. Other browsers put it between `info` and `warn`, putting it closer to `info` than `debug`.

But that’s not a strong opinion. The current level `.log()` is aliased to does not seem to be a problem for anyone.

lib/loglevel.js
```diff
@@ -34,54 +31,17 @@
...
     // Wherever possible we want to bind, not wrap, to preserve stack traces
-    function realMethod(methodName) {
-        if (methodName === 'debug') {
-            methodName = 'log';
+    function defaultMethodFactory(methodName, _level, _loggerName) {
+        if (typeof console !== undefinedType) {
+            var consoleMethod = console[methodName] || console.log;
+            if (typeof consoleMethod === "function") {
+                return consoleMethod.bind(console);
+            }
```
pimterry, line 41: Nit: imo this'd be clearer if you flip the `if` into a guard condition:

```js
if (typeof console === undefinedType) {
  return noop;
}

...
```

dist/loglevel.min.js
```diff
@@ -1,3 +1,3 @@
 /*! loglevel - v1.9.2 - https://github.com/pimterry/loglevel - (c) 2024 Tim Perry - licensed MIT */
```
pimterry, line 1: Doesn't really matter on a branch like this, but up until now I've only been updating the `dist/` files at the point of actual releases, and other changes have just been applied to `lib/` by itself. There's no strong case for it though beyond convention I guess.

In fact, the only reason `dist` is really committed at all is because we used to use github as a CDN, but nowadays unpkg can easily serve the official release content out of the npm package, so we could git-ignore `dist` entirely! (Keep building & publishing it, but no need to version or store it here at all). Not critical for this PR, but lets look at it as part of v2.

pimterry (approved):
> The only change I might still suggest is making log an alias for info instead of debug in order to work more like browser dev tools

Yep, that makes sense for me! As in the README, this was intended just as a little convenience method, not as a separate level, and I think that matches normal levels elsewhere so I don't think we should make it an actual loglevel in itself. Aliasing to info seems reasonable though.

Happy to merge this now and leave it for a separate PR, or include it here if you prefer, up to you.

lib/loglevel.js
Mr0grog: So, then we need to return the `noop` in two places. I think you’re basically recommending the logic that I previously had here? 4710d8a5f90f9156869e91a8ff19f3d6115092c8

I was trying to reduce duplication, and this read much clearer to me, but I can put the old logic back.

dist/loglevel.min.js
Mr0grog: Yeah, I assumed this was your intention, based on how that directory has been managed in the past. I updated the `dist` folder here with that in mind, actually: this is not going into the `main` branch (so it doesn't interrupt anyone using GitHub’s default branch as a source to load from) but having a build available allows people to more easily test and give feedback on on all the breaking changes and philosophical changes landing in v2. But I can pull it back out if you’d prefer.

Mr0grog:
> Aliasing to info seems reasonable though.
> 
> Happy to merge this now and leave it for a separate PR, or include it here if you prefer, up to you.

@pimterry I went ahead and made that change, so feel free to merge.

Re: the two issues you called out inline, please see my replies. Happy to make changes if you want.

dist/loglevel.min.js
pimterry: Happy to leave it for now for easy testing, but I think it would be good to drop `dist` before the final v2 release.

lib/loglevel.js
pimterry: Fair enough. I'd missed the fall-through case there. Personally I think this way round is mildly less clear than the cases, but it's not a big deal! Happy to merge as is, we can tweak and tune endlessly if we want to later on.

PR #204, opened 2025-02-25
feat: Add hierarchical category

wayfarer3130:
This change adds hierarchical categories, which the level inherits from, rather than from the root level.  This adds methods:
clearInheritedLevel
getInheritedLevel

and invokes them such that the getLevel will return updates on the parent level at the setLevel value.


pimterry:
Hi @wayfarer3130, thanks for contributing! This looks promising, but unfortunately work on the main branch is currently effectively paused, as we recently starting work on a major revamp on the `v2` branch.

@Mr0grog was working on that, and we had already discussed (but haven't started implementing) hierarchical logging there, so this would fit right in. There's some background in here: https://github.com/pimterry/loglevel/issues/119#issuecomment-2521460559.

@Mr0grog how're you doing? We haven't touched this recently, but it might be nice to keep things moving if you have some time, and my schedule is slowly become a bit easier to manage now too. Is the v2 branch in a place where @wayfarer3130 could potentially open a PR to add hierarchical logger support there? I think the general approach we'd discussed was:

* Maintaining live relationships between loggers and parents - until a logger has a level set explicitly, it should always follow the parents level, including future changes. Once it has a level is set explicitly it should become independent.
* Supporting multi-level relationships (presumably via .getLogger().getLogger() as here? Seems sensible enough to me).

I'm not sure we need extra get/clearInheritedLevel methods for this, but I'm open to discussion if we need those for a specific use case. In general the important thing is just pinning down the core behaviours that are hard to change later, and then we can tweak and tune and add more API later on after v2 is finished and we see how well it works in the real world.

wayfarer3130:
@pimterry 
> I'm not sure we need extra get/clearInheritedLevel methods for this, but I'm open to discussion if we need those for a specific use case. In general the important thing is just pinning down the core behaviours that are hard to change later, and then we can tweak and tune and add more API later on after v2 is finished and we see how well it works in the real world.

The get/clear levels are really an internal detail intended to be private, which works fine in typescript.  They are intended to allow resetting the inheritted values correctly without having to lookup the parent inheritance all the time, and without needing to reset the world.

What is the time frame for having a release build of the 2.0 version?  I'm trying to decide whether to go with another library or not.  Finding a library that works everywhere (client/server/lamdba function/bun) and has hierarchical categories is proving to be difficult and I have about 6 OSS projects that I was hoping to choose a logging library for, all of which run both client and server side.

Mr0grog:
> @Mr0grog how're you doing? We haven't touched this recently, but it might be nice to keep things moving if you have some time, and my schedule is slowly become a bit easier to manage now too.

Hey! Sorry for the radio silence — some personal stuff came up at the end of December when I thought I was going to get this done, and a lot of other stuff has kept me pulled away from it in January/February. I should have checked in about my status earlier. I definitely want to get back to this, but I am not sure if/when I will have much time for at least the next couple weeks. Leaving it in a halfway state like this is the last thing I’d wanted to do. 😓 

> Is the v2 branch in a place where @wayfarer3130 could potentially open a PR to add hierarchical logger support there?

The next thing I’d planned to do was the big switch-over into ESM/TS (and an accompanying revamp of tests and build tooling, since the current setup will immediately cease to work). That is the one big-and-can’t-be-made-much-smaller chunk of work that might be in the way.

We could definitely land hierarchical logging on the v2 branch *before* that, instead, since the ESM/TS stuff is really just a syntax change for the library — the main work is revamping test & build tooling.

I am fine if someone wants to jump in on some part of this! I’m sorry for having turned into a bit of a blocker instead of an enabler here.

> I think the general approach we'd discussed was…

Yep, that summary jives with what I was thinking. The one additional bit of sugar I thought could be nice would be something like:

```js
logger1 = loglevel.getLogger("thing1").getLogger("thing1");
logger2 = loglevel.getLogger("thing1:thing2");  // or maybe "thing1.thing2"? I think `:` is the syntax debug uses.
logger1 === logger2
```

>> I'm not sure we need extra get/clearInheritedLevel methods for this
>
> The get/clear levels are really an internal detail intended to be private, which works fine in typescript.

There are probably a lot of ways to implement. I think it would be nice to keep this library’s current behavior, where private stuff really is private and inaccessible from outside the logger module. Barring that, I think it would be good to prefix with `_`. Not all users will be TypeScript users, and having a clearer hint about privacy is good in regular old JS where a compiler is not going to yell at you.

> What is the time frame for having a release build of the 2.0 version?

For my part of this work, I’m not sure I will really have time to dig in and get things done for at least another couple weeks (more detail above). I can’t speak for @pimterry as the owner who has to review and approve all this stuff, though!

> Finding a library that works everywhere (client/server/lamdba function/bun) and has hierarchical categories is proving to be difficult and I have about 6 OSS projects that I was hoping to choose a logging library for, all of which run both client and server side.

Yeah, this is still my favorite JS logging tool (*especially* for stuff in browsers, where size matters more) for similar reasons, and that’s why I’d started this v2 work. I feel your frustration with the timing.

wayfarer3130:
> 
> Yep, that summary jives with what I was thinking. The one additional bit of sugar I thought could be nice would be something like:
> 
> ```js
> logger1 = loglevel.getLogger("thing1").getLogger("thing1");
> logger2 = loglevel.getLogger("thing1:thing2");  // or maybe "thing1.thing2"? I think `:` is the syntax debug uses.
> logger1 === logger2
> ```
 
Yes, that is the current behaviour, with dots, not colons, although the fact that the parent can be a symbol means one has to manually set inheritance.

> > > I'm not sure we need extra get/clearInheritedLevel methods for this
> > 
> > 
> > The get/clear levels are really an internal detail intended to be private, which works fine in typescript.
> 
> There are probably a lot of ways to implement. I think it would be nice to keep this library’s current behavior, where private stuff really is private and inaccessible from outside the logger module. Barring that, I think it would be good to prefix with `_`. Not all users will be TypeScript users, and having a clearer hint about privacy is good in regular old JS where a compiler is not going to yell at you.

Having private stuff private means there have to be external methods to call/update some of the values when the state transition happens.  I did in fact leave the stuff as private values, just provided some methods to access a bit of it.  The getInheritedLevel could be made private as that one isn't called externally.  However, making it all private and inaccessible means it isn't possible to write the changes externally, which is what I'd planned on doing originally (actually, that is why I ended up submitting a PR as I couldn't get access to it externally).

> 
> > What is the time frame for having a release build of the 2.0 version?
> 
> For my part of this work, I’m not sure I will really have time to dig in and get things done for at least another couple weeks (more detail above). I can’t speak for @pimterry as the owner who has to review and approve all this stuff, though!

Probably I can live with using it now as long as there is actually a commitment to get the hierarchical logging included in the next couple of months with a release of 2.0.

I'm hoping that in the 2.0 version it will be possible to inject alternate appenders - even if you only get one per log tree it is easy enough to write a splitting appender.  That way it can be used for both audit logs and debug/file logs, with various output schemes.

wayfarer3130:
If you let me know where the 2.0 branch is, I could add this to the 2.0 branch

Mr0grog:
> I'm hoping that in the 2.0 version it will be possible to inject alternate appenders

I think designing a more intentional plugin API is another big goal for 2.0, so if you have use cases or design ideas, putting them in #119 would be great.

For built-in support, my understanding is that we want to keep it small and focused on `console`, but that it should be as simple as possible to change that in a plug-in.

> If you let me know where the 2.0 branch is

It’s [`v2`](https://github.com/pimterry/loglevel/tree/v2).

pimterry:
> Hey! Sorry for the radio silence — some personal stuff came up at the end of December when I thought I was going to get this done, and a lot of other stuff has kept me pulled away from it in January/February.

No worries, of course! We all have a lot going on and it's hard to find time for open-source like this, I totally understand.

> What is the time frame for having a release build of the 2.0 version?

> For my part of this work, I’m not sure I will really have time to dig in and get things done for at least another couple weeks (more detail above). I can’t speak for @pimterry as the owner who has to review and approve all this stuff, though!

> Probably I can live with using it now as long as there is actually a commitment to get the hierarchical logging included in the next couple of months with a release of 2.0.

There will certainly be hierarchical logging in v2. When exactly it will be released is hard to say, since that'll depend on the other changes we want to include but a few months is plausible I think (and I do have a bit more time nowadays for reviewing, so hopefully I won't be a blocker there). In the meantime of course you can fork and build your own version directly from the v2 branch, and then upgrade to the final version once that's released.

> However, making it all private and inaccessible means it isn't possible to write the changes externally, which is what I'd planned on doing originally (actually, that is why I ended up submitting a PR as I couldn't get access to it externally).

Not being able to mess with internals like this is generally preferable. The problem is that this library is _very_ widely used (currently 9.4 million installs a week!) and anything that's publicly available as part of it will be used by people in every way you can imagine, and then it's a public API that we have to support and maintain forever (even if you tell them not to and say its not supported, it will still create bug reports & support hassle and it's just nice to not break people's code unexpectedly).

When people start depending on internals and you just can't change them any more, you quickly run into all sorts of problems with ongoing development and evolution in future (I'm one of the Node.js maintainers - Node has this problem in spades, making it nearly impossible to refactor large parts of the codebase).

Conclusion: we can't expose any public fields or methods that we don't intend users to actively use, so things do need to be really private and internal wherever possible, not just hidden from the types. If there are tricky cases though, do feel free to just open a draft PR with them public regardless, and we can discuss the options from there.

> I'm hoping that in the 2.0 version it will be possible to inject alternate appenders

We don't have a final design anywhere I think, but yes I do think that should be well supported - it's possible with the plugin API today really, your plugin gets called with every log message, and you can decide whether to call through to console.log or pass the message elsewhere as you'd like. Feel free to open an issue with suggestions with you have opinions on this (@Mr0grog I think we should try to split the plan out into separate issues from just #119 as we continue the planning now, just to try and keep the different threads manageable!).

> The one additional bit of sugar I thought could be nice would be something like: ["child:grandchild"]

Imo, if we can and unless people feel strongly, it would be nice to avoid logic powered by special symbols within strings, since there will inevitably be somebody who uses them unintentionally or because they're part of an existing naming scheme or something, and then gets strange confusing behaviour. In-band signalling is nice to avoid if we can.

What do you think about handling this like so:

```js
loglevel.getLogger("child", "grandchild")
```

At the cost of 3 extra characters (or 2, ignoring whitespace), by splitting into two params we make it completely explicit what's going on, we don't have to parse strings at all, and if people want to use dots or colons or whatever within their logger names they're free to do so. Would that work?

wayfarer3130:
> What do you think about handling this like so:
>
> loglevel.getLogger("child", "grandchild")
> At the cost of 3 extra characters (or 2, ignoring whitespace), by splitting into two params we make it completely explicit what's > going on, we don't have to parse strings at all, and if people want to use dots or colons or whatever within their logger names they're free to do so. Would that work?

Yes, I like that a lot better.  It also allows symbols at any level, or numbers even.

And
loglevel.getLogger("child").getLogger("grandchild") === loglevel.getLogger("child").getLogger("grandchild")

With typescript, I will use the ... syntax for getting the list array.

For 1.x, would you be amenable to adding JUST the child getLogger capability mapping the child levels to the "child.grandchild" syntax for now, without the inheritance?  That would allow for a better transition for the syntax.  In specific, I would implement just:
root getLogger(a,b,c) => getLogger(a).getLogger(b).getLogger(c)
and childLogger.getLogger(name) => root.getLogger(`${childLogger.name}.${childName}`)

Otherwise I will implement that as an add on feature inside one of my libraries.

pimterry:
> For 1.x, would you be amenable to adding JUST the child getLogger capability mapping the child levels to the "child.grandchild" syntax for now, without the inheritance? That would allow for a better transition for the syntax.

I don't think so - I'd really like to avoid adding any unnecessary complexity or making more changes within v1 right now. I think it's best to consider it frozen from now on for anything but critical bug fixes.

If you're looking for this heirarchy functionality ASAP, as soon as it's been added to v2 (PRs welcome!) then you can just use the v2 branch directly, either installing that from GitHub or forking and publishing that as a standalone fork on npm yourself. A setup like that is likely to provide a much smoother transition towards the final v2 release than any kind of intermediate changes in v1.

wayfarer3130:
Closing in favour of a v2 change

pimterry (commented):
(wrong PR - deleted)

Mr0grog:
>> The one additional bit of sugar I thought could be nice would be something like: ["child:grandchild"]
> 
> Imo, if we can and unless people feel strongly, it would be nice to avoid logic powered by special symbols within strings, since there will inevitably be somebody who uses them unintentionally or because they're part of an existing naming scheme or something, and then gets strange confusing behaviour. In-band signalling is nice to avoid if we can.

Catching up here, can move this to #205 (haven’t read it yet), but as long as we plan to support external config via env vars/querystrings/local storage/cookies/whatever, there’s going to have to be a format for this involving some kind of special symbols within strings. Obviously we can invent some kind of escaping scheme for persistence, but I think it’s just much clearer and simpler for users if they way they specify hierarchy in, say, an environment variable, matches how they do it in code, so we may as well decide that here.

I can appreciate that people might be using dots or colons or whatever in their current names, and this would present a breaking change for them, but that is kind of the point of doing this as v2 instead of v1.x.

>> What do you think about handling this like so:
>>
>> ```js
>> loglevel.getLogger("child", "grandchild")
>> ```
>
> Yes, I like that a lot better. It also allows symbols at any level, or numbers even.

I think this works great, too, and is a good way to preserve support for symbols. I am a little wary of supporting other primitive types (e.g. numbers), but that is probably a different discussion.

The one thing I’d like to note here is that I think I suggested elsewhere (probably back in #119) supporting an options object in `getLogger(name: string, options?: Options)`. That isn’t necessarily precluded by allowing a variable number of name arguments, but it does make argument handling and typing more complicated.

PR #205, opened 2025-03-04
feat: v2 hierarchical categories

wayfarer3130:
Add hierarchical categories in loglevel v2


lib/loglevel.js
```diff
@@ -189,68 +189,105 @@
       self.methodFactory = factory || defaultMethodFactory;
 
       self.getLevel = function () {
-          if (userLevel != null) {
-            return userLevel;
-          } else if (defaultLevel != null) {
-            return defaultLevel;
-          } else {
-            return inheritedLevel;
+          if (userLevel !== null) {
```
wayfarer3130: This fixes a bug where user level 0 wasn't applying.

lib/loglevel.js
```diff
@@ -78,19 +78,19 @@
        * in this child yet).
        * @type {number}
        */
-      var inheritedLevel;
+      var inheritedLevel = null;
```
wayfarer3130: These null changes fix a bug causing the levels to not apply correctly.

lib/loglevel.js
wayfarer3130, line 1: @pimterry  - For getLogger, I'd like to make it take a `...categories` argument, like this:
`getLogger(...categories)`
instead of a name argument, but it still needs to be compatible with the old name arguments, and the get loggers still needs to get a single logger setup.  What I'm thinking is that the name object will become 
```
const isLastSymbol = typeof (categories[categories.length-1]) === 'symbol';
const name = isLastSymbol ? categories[categories.length-1]) : categories.join('.');
```

So that we preserve the existing behaviour on symbols, and use a combined name for dots.
Alternatively, I could create a combined symbol for the overall name.

Either way, the parent logger would become `getLogger(categories.slice(0,categories.length-1))`  which causes some issues if anyone tries getting the `category0.category1` categories.
If it was a string containing a . then a split could be done, preserving full compatibility.

Let me know which would work better for backwards compatibility.

wayfarer3130: I could also just add getCategory instead of getLogger - but the name in the overall list still would need to be defined.

.gitignore
```diff
@@ -22,3 +24,6 @@ npm-debug.log
 
 _SpecRunner.html
 .grunt
+
+dist/
+*.code-workspace
```
pimterry: This line is in here twice

.gitignore
```diff
@@ -22,3 +24,6 @@ npm-debug.log
 
 _SpecRunner.html
 .grunt
+
+dist/
```
pimterry: I see you've dropped this and the `dist` files. @Mr0grog and I had some discussion around this [here](https://github.com/pimterry/loglevel/pull/203#discussion_r1896010585), with the conclusion it was useful to keep them until the final release of v2 for easier testing.

I'm open to debate, but I think that's a reasonable plan for now, so we should undo this. You can commit the rebuilt versions or we can update them after the PR is done, up to you.

README.md
```diff
@@ -243,6 +243,46 @@ Like the root logger, other loggers can have their logging level saved. If a log
...
+
+This gets you a new logger object which inherits the parent logger's log level, being named `parentLogger.childLogger`.  That allows using `setLevel` on the parent logger and having that apply to the child logger.  Suppose each class has it's own logger in the form `app` with child `ui`, `controller`, `model`, `util` etc followed by the class name.  Now, you might set the `app` to `debug`, but you don't care about the util logging, so you set `app.util` to `warn`.  That is two different levels to set, and might apply to a large number of classes.  You might also decide you don't care about `app.model.NoisyClass`, so you can set that one to warn, but still leave the other model ones active.  Or conversely, set the `app.model` to `warn` and then a model class you do care `app.model.ImportantClass` to debug.  
+
+Example usage _(using CommonJS modules, but you could do the same with any module system):_
+
+```javascript
+// In module-one.js:
+var loglevel = require("loglevel");
+loglevel.setLevel("error");
+var logApp = .getLogger("app");
```
pimterry: This is missing a `loglevel` before the `.`

README.md
```diff
@@ -243,6 +243,46 @@ Like the root logger, other loggers can have their logging level saved. If a log
...
 
+#### Hierarchical Loggers `log.getLogger(parentLoggerName).getLogger(childLoggerName)`
+
+This gets you a new logger object which inherits the parent logger's log level, being named `parentLogger.childLogger`.  That allows using `setLevel` on the parent logger and having that apply to the child logger.  Suppose each class has it's own logger in the form `app` with child `ui`, `controller`, `model`, `util` etc followed by the class name.  Now, you might set the `app` to `debug`, but you don't care about the util logging, so you set `app.util` to `warn`.  That is two different levels to set, and might apply to a large number of classes.  You might also decide you don't care about `app.model.NoisyClass`, so you can set that one to warn, but still leave the other model ones active.  Or conversely, set the `app.model` to `warn` and then a model class you do care `app.model.ImportantClass` to debug.  
+
+Example usage _(using CommonJS modules, but you could do the same with any module system):_
+
+```javascript
+// In module-one.js:
+var loglevel = require("loglevel");
```
pimterry: We should fully move to modern syntax in the docs for the new release (we'll have some backward compat of course, but modern formatting for the docs themselves would be best) so lets do everything ESM + `const`/`let` instead of `var`.

wayfarer3130:
@pimterry - I added the hierarchical loggers and the `getLogger("parent").getLogger("child") === getLogger("parent","child")`
with the inheritance of the default log level.

test/node-integration.js
```diff
@@ -44,4 +44,33 @@ describe("loglevel included via node", function () {
...
+        expect(logger1.getLevel()).toBe(log.levels.INFO);
+        var logger12 = logger1.getLogger('two');
+        expect(logger12.getLevel()).toBe(log.levels.INFO);
+        log.setLevel("warn");
+
+        expect(log.getLevel()).toBe(log.levels.WARN);
+        expect(logger12.getLevel()).toBe(log.levels.WARN);
+    });
+
+    it("supports child logger inheritance", function() {
```
pimterry: These two tests have the same name

lib/loglevel.js
```diff
@@ -189,68 +195,108 @@
...
+          return normalizeLevel(
+              defaultLogger ? defaultLogger.getLevel() : "WARN" );
       };
 
-      self.setDefaultLevel = function (level) {
-          defaultLevel = normalizeLevel(level);
-          if (!getPersistedLevel()) {
-              self.setLevel(level, false);
+      /** The parentKey is the name of the parent this inherits from */
+      self.parentKey = undefined;
```
pimterry: Loggers here store separate variables for `name`, `categoriesParam`, `parentKey` and `categories` as a separate public field, and `newCategories` & `childCategories` fields added in getLogger. All of those are fundamentally derived data of a single thing.

I think we only need one of these: Logger should take a single `keyHierarchy` that's an `Array<string | Symbol>` (no separate name + hierarchy).

That is effectively `categoriesParam`. `parentKey` is then `keyHierarchy.slice(0, -1)`, name is `keyHierarchy.join('.')`. `categories`, `newCategories` and `childCategories` can go.

I think there's a general design simplification that will follow if we head along these lines - I'll try to expand on that in the comments below.

lib/loglevel.js
```diff
@@ -189,68 +195,108 @@
...
+              return defaultLogger.getLogger(self.parentKey).getLevel();
           }
-
-          // NOTE: in v2, this should call rebuild(), which updates children.
-          return replaceLoggingMethods.call(self);
+          if( !defaultLogger || defaultLogger===self ) {
+              return normalizeLevel("WARN");
+          }
+          return normalizeLevel(
+              defaultLogger ? defaultLogger.getLevel() : "WARN" );
```
pimterry: I think this method can be simplified: `defaultLogger` is always set, and with `parentKey` gone that bit becomes `defaultLogger.getLogger(...keyHierarchy.slice(0, -1)).getLevel()`, which should work for all loggers except the root so we don't need a parentKey check. We don't need normalizeLevel either. Could be something like:

```js
if (defaultLogger === self) {
  return self.levels.WARN;
} else {
  return defaultLogger.getLogger(...keyHierarchy.slice(0, -1)).getLevel();
}

File: lib/loglevel.js [hunk 10]
```diff
@@ -261,18 +307,33 @@
...
-        if ((typeof name !== "symbol" && typeof name !== "string") || name === "") {
-            throw new TypeError("You must supply a name when creating a logger.");
+    defaultLogger.getLogger = function getLogger(_singleName) {
+        if( !arguments.length ) {
+            throw new TypeError('No arguments provided to getLogger');
+        }
+        var categories = Array.from(arguments);
+        var lastCategory = categories[categories.length-1];
+        var isSymbolName = typeof lastCategory === 'symbol';
+        var name = categories.length===1 ? lastCategory : categories.join('.');
```
pimterry: We shouldn't use dot-joined naming for the internal name of loggers. That results in weird things like `getLogger("a", "b") === getLogger("a.b")` which is what I was trying to avoid [in my previous comment](https://github.com/pimterry/loglevel/pull/204#issuecomment-2690474817). That's also why symbols aren't supported here.

I think the main reason this is here is because we need a string name for `_loggersByName`, is that right? We can fix that. Instead, we can have `_loggersByHierarchy`, where each level is a name part, so a populated data structure would look like:

```js
_loggersByHierarchy = {
  "app": {
    "ui": {
      "class1": <logger instance>
    }
  }
};
```

That gives us much better visibility into the structure later on. It also makes `_clearInheritedLevel` much nicer, because we can specify a specific parent and then easily get all exactly the affected children instead of having to update every single logger every time.

Does that make sense?

lib/loglevel.js
```diff
@@ -261,18 +307,33 @@
 
     defaultLogger = new Logger();
 
-    defaultLogger.getLogger = function getLogger(name) {
-        if ((typeof name !== "symbol" && typeof name !== "string") || name === "") {
-            throw new TypeError("You must supply a name when creating a logger.");
+    defaultLogger.getLogger = function getLogger(_singleName) {
+        if( !arguments.length ) {
```
pimterry: For consistency and simplicity elsewhere, it'll actually be nicer to allow this, such that `logger.getLogger()` returns `logger`. It'll make the model & code cleaner to avoid the special case, and means that things like `defaultLogger.getLogger(keyHierarchy.slice(0, -1))` work correctly for the first level of the hierarchy.

lib/loglevel.js
```diff
@@ -189,68 +195,108 @@
...
-          userLevel = null;
-          clearPersistedLevel();
-          replaceLoggingMethods.call(self);
+      self.getLogger = function(name) {
+          var childCategories = Array.from(arguments);
+          var newCategories = this.categories ? this.categories.concat(childCategories) : childCategories.slice();
+          var childLogger = defaultLogger.getLogger.apply(defaultLogger,newCategories);
+          childLogger.childCategories = childCategories;
+          childLogger.newCategories =newCategories;
+          return childLogger;
```
pimterry: Similarly, I think this whole method just becomes `defaultLogger.getLogger(...keyHierarchy, ...arguments)`.

pimterry (commented):
I've added quite a few comments. This is a great start but there's a few ways we can simplify this. I've tried to cover the main points here so you get an idea - the main thing is that I don't think we need also the extra fields and edge cases.

README.md
wayfarer3130: Whenever the build gets fixed so it doesn't die on that I can change to it.

lib/.jshintrc
```diff
@@ -1,21 +1,21 @@
...
+    "boss": true,
+    "eqnull": true,
+    "notypeof": true,
+    "globals": {
+        "console": false,
+        "exports": false,
+        "define": false,
+        "module": false,
+        "window": false
+    }
```
Mr0grog: Minor nit: this indentation change [contravenes `.editorconfig`](https://github.com/pimterry/loglevel/blob/7a4b1d963154874f6035338b46d269c40e5d2761/.editorconfig#L20-L21). Either this or that should change so they match up.

lib/loglevel.js
```diff
@@ -99,57 +99,63 @@
...
+                window.localStorage[storageKey] = levelName;
+                return;
+            } catch (ignore) {}
+
+            // Use session cookie as fallback
+            try {
+                window.document.cookie =
+                    encodeURIComponent(storageKey) + "=" + levelName + ";";
+            } catch (ignore) {}
+        }
```
Mr0grog: I think the indentation here is messed up? It seems like this is also causing unnecessary indentation changes to all the code after this, making the diff a little harder to follow (lots of things getting marked as changed that are not, and getting interleaved in the diff with new code when it otherwise would not be).

lib/loglevel.js
```diff
@@ -189,68 +195,108 @@
...
+              inheritedLevel = self.getInheritedLevel();
+          }
+          return inheritedLevel;
       };
 
-      self.setLevel = function (level, persist) {
-          userLevel = normalizeLevel(level);
-          if (persist !== false) {  // defaults to true
-              persistLevelIfPossible(userLevel);
+      self.getInheritedLevel = function() {
```
Mr0grog: Is it important that this is public? This seems like an implementation detail of `getLevel()`. In keeping with [@pimterry’s earlier comments](https://github.com/pimterry/loglevel/pull/204#issuecomment-2694383418), I think it would be good to keep this private for now. It can always be made public later if there are strong use cases.

lib/loglevel.js
```diff
@@ -189,68 +195,108 @@
...
       };
 
-      self.setDefaultLevel = function (level) {
-          defaultLevel = normalizeLevel(level);
-          if (!getPersistedLevel()) {
-              self.setLevel(level, false);
+      /** The parentKey is the name of the parent this inherits from */
+      self.parentKey = undefined;
+
+      self._clearInheritedLevel = function () {
```
Mr0grog: Same as `getInheritedLevel()` above about this being publicly accessible.

test/.jshintrc
```diff
@@ -1,34 +1,34 @@
...
+        "context": false,
+        "describe": false,
+        "xdescribe": false,
+        "expect": false,
+        "it": false,
+        "jasmine": false,
+        "waitsFor": false,
+        "runs": false,
+        "Symbol": false
+    }
```
Mr0grog: Same as above with respect to indentation and `.editorconfig`.

test/multiple-logger-test.js
```diff
@@ -24,9 +24,10 @@ define(['test/test-helpers'], function(testHelpers) {
                 expect(newLogger.methodFactory).toBeDefined();
             });
 
-            it("returns loggers without `getLogger()` and `noConflict()`", function(log) {
+            it("returns loggers without `noConflict()`", function(log) {
                 var newLogger = log.getLogger("newLogger");
-                expect(newLogger.getLogger).toBeUndefined();
+                // Now expect getLogger to be defined for the child
+                // expect(newLogger.getLogger).toBeUndefined();
```
Mr0grog: I think you can just remove this entirely; it’s totally irrelevant for v2. Is there an important reason to preserve the old behavior as a comment here?

Mr0grog (commented):
I don’t want to step on @pimterry’s toes in this review — these are just a few minor notes about syntax and public/private stuff.

lib/loglevel.js
wayfarer3130: According to the editor config, it is supposed to be 4 spaces indent, which is what this is.  It was previously a mix of 4 and 2 spaces.  I'm fine/better off with 2, just saying that the configuration says 4.   Note that is NOT true for JSON, which IS 2 spaces.

wayfarer3130:
@pimterry @Mr0grog 
I've updated this to address the comments.  
* Use rebuild to update both the method and inherited level
* getLogger() returns the logger itself, at whatever level that happens to be, that is, log.getLogger()===log
* Inherit methodFactory from parent, and children can inherit from any log instance after a rebuild
* No manual rebuild is needed
* No storage on symbols, but they can inherit from parent symbols
* Single method for getLogger for both default and child loggers.

It needs docs updated, but I'm not going to complete that until the code is approved, then I will change just the docs.

lib/loglevel.js
Mr0grog: Right. I was referring to the fact that ([at the time of this comment](https://github.com/pimterry/loglevel/blob/919402bf66132be5b32fd035e78351452d74af35/lib/loglevel.js#L114-L124)), line 115 was indented 6 from line 114 instead of 4, and then when dedenting on line 117 and later, everything was off by 2 (outdented 4 instead of 6). But it looks like you fixed that in 3b00cad2688ffc667dea164b5108bbcb3128ec37 🙂

README.md
```diff
@@ -86,19 +85,28 @@ loglevel is written as a UMD module, with a single object exported. Unfortunatel
...
+log.warn("module-static");
 ```
 
 For some tools though, it might better to wildcard import the whole object:
 
 ```javascript
-import * as log from 'loglevel';
-log.warn("module-tastic");
+import * as log from "loglevel";
+log.warn("module-static");
```
pimterry: Haha, this isn't actually a typo, it's intended as a module + [tastic](https://dictionary.cambridge.org/dictionary/english/tastic).

README.md
```diff
@@ -86,19 +85,28 @@ loglevel is written as a UMD module, with a single object exported. Unfortunatel
...
 
 There's no major difference, unless you're using TypeScript & building a loglevel plugin (in that case, see <https://github.com/pimterry/loglevel/issues/149>). In general though, just use whichever suits your environment, and everything should work out fine.
 
+### Non minified code
+
+The non-minified code can be imported in module based systems using:
+
+```javascript
+import log from "loglevel/nomin";
+log.warn("module-not-minified");
```
pimterry, line 107: I don't think we want to add a new entrypoint like this without separate discussion. I suspect we don't want this at all, and the default entrypoint can just always be non-minified (and was in v1 too). But regardless it's a separate question, let's worry about minification & export format changes later.

lib/loglevel.js
```diff
@@ -157,116 +165,119 @@
...
+
+        self.getLogger = function (...childCategories) {
+            if (!childCategories.length) {
+                return self;
+            }
+            let newCategories = this.categories ? [...this.categories] : [];
+            return childCategories.reduce((logger, category) => {
+                let childLogger = logger
+                    .getLoggers()
+                    .find((logger) => logger.name === category);
```
pimterry: `loggers` can be a dictionary still, and then this will just be `logger.getLoggers()[category]`. Not a big deal, but it'd be a bit faster & simpler that way. That might make getLoggers nicer too, I think exposing a dictionary there is probably what people want. We can still iterate over it with `Object.values(loggers)` for rebuild. 

I don't think there's any other specific advantage to making it an array, although I'd be interested if so.

lib/loglevel.js
```diff
@@ -157,116 +165,119 @@
...
+                        throw new Error('Category names must be a symbol or string, but is a '+typeof category);
+                    }
+                    childLogger = new Logger(category, self.methodFactory, logger);
+                    loggers.push(childLogger);
+                }
+                return childLogger;
+            }, self);
+        };
+
+        self.getLoggers = function () {
```
pimterry: Maybe new API should be `getChildLoggers`? Just to be a bit clearer

package.json
```diff
@@ -21,15 +21,19 @@
   },
   "license": "MIT",
   "main": "lib/loglevel.js",
+  "exports": {
+    "./nomin": "./dist/loglevel.js",
+    "default": "./dist/loglevel.min.js"
+  },
```
pimterry: We do need to explore `exports`, especially for ESM support etc, but I don't think we need to do that here, we can leave that out of this PR.

pimterry (commented):
Nice work @wayfarer3130! I like the approach. A couple more comments but just on small details now, I think this is nearly there.

There are other details we could tweak, but we can fix those up later (I agree the indentation is currently a mess for example, we should standardize on something there - personally I'd lean to 4 spaces everywhere but whatever people prefer).

This is mergeable though imo once the two `loggers` API details here are covered and the exports changes are pulled out.

lib/loglevel.js
Ryuno-Ki: Dictionary feels so Pythonic :innocent: 
(JavaScript has Objects)

wayfarer3130:
Made the changes you requested

pimterry (approved):

pimterry:
Nice stuff, thanks @wayfarer3130! Merged :+1:

PR #209, opened 2026-01-21
feat: typescript version

wayfarer3130:
This is a potential version for typescript - the api is compatible, but I haven't used much extra in terms of the typescript versioning of any ability to over-ride a class instance.

Note the lib/loglevel.js javascript version is GENERATED now from src/loglevel.ts.  That is one of the things to discuss about what level of backwards compatibility this should have and how much new features should be taken advantage of.
Please provide comments about overall goals for a v2 - we can discuss the nitpicky details once we have the high level agreement on the general pattern.


wayfarer3130:
@pimterry - this is a potential version converted to typescript, mostly as a stalking horse to talk about to see if we can figure out the right changes for a v2 release.

.github/workflows/ci.yml
```diff
@@ -1,12 +1,10 @@
 name: Continuous Integration
 
 on:
-  pull_request: {}
-  push:
-    branches:
-      - main
```
Ryuno-Ki: Removing this would make it run for pushes to any branch.

.gitignore
```diff
@@ -12,6 +12,10 @@ lib-cov
 *.pid
 *.gz
 *.tgz
+*.code-workspace
+package-lock.json
```
Ryuno-Ki, line 16: This file should be tracked in version control.

.gitignore
```diff
@@ -12,6 +12,10 @@ lib-cov
...
+bun.lock
```
Ryuno-Ki, line 17: This file should be tracked in version control.

lib/loglevel.js
```diff
@@ -17,11 +17,8 @@
     "use strict";
 
     // Slightly dubious tricks to cut down minimized file size
-    var noop = function() {};
-    var undefinedType = "undefined";
-    var isIE = (typeof window !== undefinedType) && (typeof window.navigator !== undefinedType) && (
-        /Trident\/|MSIE /.test(window.navigator.userAgent)
-    );
+    const noop = function() {};
```
Ryuno-Ki: If you want to be consistent:

```suggestion
    const noop = function () {};
```

lib/loglevel.js
```diff
@@ -31,57 +28,19 @@
...
-        }
-        if (console.trace) console.trace();
-    }
-
     // Build the best logging method possible for this env
     // Wherever possible we want to bind, not wrap, to preserve stack traces
-    function realMethod(methodName) {
-        if (methodName === 'debug') {
-            methodName = 'log';
+    function defaultMethodFactory(methodName, _level, _loggerName) {
```
Ryuno-Ki:
```suggestion
    function defaultMethodFactory (methodName, _level, _loggerName) {
```

lib/loglevel.js
```diff
@@ -31,57 +28,19 @@
...
+        if (typeof console !== undefinedType) {
+            var consoleMethod = console[methodName] || console.log;
```
Ryuno-Ki:
```suggestion
            const consoleMethod = console[methodName] || console.log;
```

lib/loglevel.js
```diff
@@ -98,218 +57,255 @@
...
+    function Logger(name, factory, parent) {
+        // Private instance variables.
+        const self = this;
+        /**
+         * The level inherited from a parent logger (or a global default). We
+         * cache this here rather than delegating to the parent so that it stays
+         * in sync with the actual logging methods that we have installed (the
+         * parent could change levels but we might not have rebuilt the loggers
+         * in this child yet).
+         * @type {number}
```
Ryuno-Ki: Doesn't this conflict with the initialisation?

```suggestion
         * @type {number|null}
```

lib/loglevel.js
```diff
@@ -98,218 +57,255 @@
...
+            if (persist !== false) {
+                // defaults to true
+                persistLevelIfPossible(userLevel);
+            }
+            return self.rebuild();
+        };
+
+        self.setDefaultLevel = function (level) {
+            defaultLevel = normalizeLevel(level);
+            if (getPersistedLevel()===undefined) {
```
Ryuno-Ki: Nitpick:

```suggestion
            if (getPersistedLevel() === undefined) {
```

lib/loglevel.js
```diff
@@ -98,218 +57,255 @@
...
+        };
+
+        self.disableAll = function (persist) {
+            self.setLevel(self.levels.SILENT, persist);
+        };
+
+        self.rebuild = function () {
+            inheritedLevel = null;
+            const result = replaceLoggingMethods.call(self);
+            if( self.methodFactory && self.methodFactory===(factory || defaultMethodFactory)) {
```
Ryuno-Ki:
```suggestion
            if (self.methodFactory && self.methodFactory === (factory || defaultMethodFactory)) {
```

test/console-fallback-test.js
```diff
@@ -15,60 +15,31 @@ function mockConsole() {
...
-                it("logging is re-enabled but does nothing when used at a blocked level", function () {
-                    log.setLevel(log.levels.WARN);
+        afterEach(function() {
+            window.console = originalConsole;
+        });
 
-                    window.console = mockConsole();
-                    log.trace("trace");
+        it("silent method calls are allowed", function() {
+            var result = log.setLevel(log.levels.SILENT);
```
Ryuno-Ki, line 28: Maybe worth to replace the `var`s here, too.

test/multiple-logger-test.js
```diff
@@ -149,6 +135,20 @@ define(['test/test-helpers'], function(testHelpers) {
               var newLogger = log.getLogger("newLogger");
               expect(newLogger).toBeAtLevel("trace");
             });
+
+            it("child loggers inherit parent", function(log) {
+                log.setLevel(0);
```
Ryuno-Ki, line 140: Hm. This appears to be an implementation detail.

Ryuno-Ki (commented):
I highlighted some areas where the spaces appear to be out of order.
Replacing `var`s can be a chore taken care of in another PR.

wayfarer3130:
> I highlighted some areas where the spaces appear to be out of order. Replacing `var`s can be a chore taken care of in another PR.

The lib/loglevel.js file is now a generated file for compatibility - if v2 doesn't need that level of compatibility, then the ts can be the primary version and we can avoid generating a lib/loglevel.js    Alterantively, we could leave lib/loglevel.ts as the primary typescript version, and add a compatible version in dist, and change the exports in package.json.

Ryuno-Ki:
Ah, so that's what you meant with TS! I was confused as I haven't seen any changes to TS files.

pimterry:
I'm very open to migrating to TS! Fully agree that would be good. It should happen on the V2 branch though, we're not migrating v1 to TS - it's frozen and there's risk for little benefit. This PR currently targets the main branch and seems to include all the other V2 changes here as well, which makes it hard to review. Can you update it to target `v2`, and only include the TS conversion?

Mr0grog:
@wayfarer3130 sorry I’ve been so absent when I thought I’d have time to work on this. Glad someone’s taking it on.

Did you maybe push the wrong commit to your PR branch? This PR appears to be the exact same as the current `v2` branch in this repo (same hash, so nothing’s been inserted into the history), and there’s no `src/loglevel.ts` file like you’re describing.

wayfarer3130:
> @wayfarer3130 sorry I’ve been so absent when I thought I’d have time to work on this. Glad someone’s taking it on.
> 
> Did you maybe push the wrong commit to your PR branch? This PR appears to be the exact same as the current `v2` branch in this repo (same hash, so nothing’s been inserted into the history), and there’s no `src/loglevel.ts` file like you’re describing.

I must have pushed it wrong somehow as it doesn't seem to have the changes I had locally.  I will push that on Monday when I'm back on the computer I was working on.
Even when I do push it, I'm mostly thinking it is a discussion point, and is extremely unlikely to be what we use for v2, just that I wanted to get people talking about what they want in v2.

lib/loglevel.js
```diff
@@ -1,334 +1,229 @@
...
-        "warn",
-        "error"
-    ];
-
-    var defaultLogger = null;
-
-    // Build the best logging method possible for this env
-    // Wherever possible we want to bind, not wrap, to preserve stack traces
-    function defaultMethodFactory(methodName, _level, _loggerName) {
+/*! loglevel - https://github.com/pimterry/loglevel - licensed MIT */
```
Ryuno-Ki, line 1: I'm not sure whether the comment can be changed into this form legally. Would it be possible to preserve the previous version?

src/loglevel.ts
```diff
@@ -0,0 +1,267 @@
...
+  methodName: LogMethodName,
+  level: number,
+  loggerName: LoggerName
+) => LoggingMethod;
+
+const logMethods: LogMethodName[] = ["trace", "debug", "info", "warn", "error"];
+
+function defaultMethodFactory(methodName: LogMethodName): LoggingMethod {
+  if (typeof console !== undefinedType) {
+    const anyConsole = console as any;
```
Ryuno-Ki, line 30: Can we prefer `unknown` over `any`?

src/loglevel.ts
```diff
@@ -0,0 +1,267 @@
...
+    }
+  }
+  return noop;
+}
+
+function replaceLoggingMethods(this: any): string | void {
+  /*jshint validthis:true */
+  const level = this.getLevel();
+
+  for (let i = 0; i < logMethods.length; i++) {
```
Ryuno-Ki, line 43: My hunch is that a [`for…of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) loop is perhaps the right way?

Ryuno-Ki: Or perhaps even a `logMethods.forEach((method, index) => {…})`?

src/loglevel.ts
```diff
@@ -0,0 +1,267 @@
...
+    const methodName = logMethods[i];
+    this[methodName] = i < level ? noop : this.methodFactory(methodName, level, this.name);
```
Ryuno-Ki, line 45: The `i` should maybe be casted here. It feels kind of confusing that `level` is an integer in this line. (It makes me think which I interpret as a smell)

src/loglevel.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  // Make `log.log` an alias to ensure compatibility with `console.*`.
+  this.log = this.info;
+
+  if (typeof console === undefinedType && level < this.levels.SILENT) {
+    return "No console available for logging";
+  }
+}
+
+function Logger(this: any, name?: LoggerName, factory?: MethodFactory, parent?: any) {
```
Ryuno-Ki, line 56: Might benefit from a return type.

src/loglevel.ts
```diff
@@ -0,0 +1,267 @@
...
+  const self = this;
+
+  // Private instance variables:
```
Ryuno-Ki, line 59: JavaScript has [private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) by now.

src/loglevel.ts
```diff
@@ -0,0 +1,267 @@
...
+  let inheritedLevel: number | null = null;
+  let defaultLevel: number | null = null;
+  let userLevel: number | null = null;
+
+  // Categories form an inherited hierarchy. Root logger has [].
+  const parentCategories: any[] = parent && parent.categories ? parent.categories : [];
```
Ryuno-Ki, line 65: Would this work, too? It's hard for me to tell what shape `parent` has here.

```suggestion
  const parentCategories: any[] = parent?.categories || [];
```

src/loglevel.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  const storageKey: string | null = hasSymbol
+    ? null
+    : parent
+      ? "loglevel:" + self.categories.map((c: any) => String(c)).join(".")
+      : "loglevel";
+
+  const loggers: any = {};
+
+  function persistLevelIfPossible(levelNum: number): void {
```
Ryuno-Ki, line 85: Might make sense to define a type for Indices like this:

```ts
// Source - https://stackoverflow.com/a/63904714
// Posted by bela53
// Retrieved 2026-02-04, License - CC BY-SA 4.0

type Indices<T extends readonly any[]> = Exclude<Partial<T>["length"], T["length"]>
```

Would then be

```suggestion
  function persistLevelIfPossible(levelNum: Indices<typeof LogMethodName[]>): void {
```

if I read it correctly.

src/loglevel.ts
```diff
@@ -0,0 +1,267 @@
...
+    const levelName = (logMethods[levelNum] || "silent").toUpperCase();
+    if (typeof window === undefinedType || !storageKey) return;
+
+    try {
+      (window as any).localStorage[storageKey] = levelName;
```
Ryuno-Ki, line 90: Is casting `window` really necessary here?
You check for `typeof window` above. Maybe there should be a general check for browser context before handling finer levels.

src/loglevel.ts
```diff
@@ -0,0 +1,267 @@
...
+    } catch (_ignore) {}
+
+    try {
+      window.document.cookie = encodeURIComponent(storageKey) + "=" + levelName + ";";
+    } catch (_ignore) {}
+  }
+
+  function getPersistedLevel(): string | void {
+    let storedLevel: any;
+    if (typeof window === undefinedType || !storageKey) return;
```
Ryuno-Ki, line 101: This starts to ask for a refactoring that handles the question of browser vs. server context generally.

package.json
```diff
@@ -30,9 +39,9 @@
     "test": "grunt test && npm run test-types",
     "test-browser": "grunt test-browser --",
     "test-node": "grunt test-node --",
-    "test-types": "tsc --noEmit ./test/type-test.ts && ts-node ./test/type-test.ts",
-    "dist": "grunt dist",
-    "dist-build": "grunt dist-build",
+    "test-types": "tsc --noEmit ./test/type-test.ts && ts-node --compilerOptions \"{\\\"module\\\":\\\"CommonJS\\\"}\" ./test/type-test.ts",
```
Ryuno-Ki, line 42: Do we really have to have this level of escaping?


```suggestion
    "test-types": "tsc --noEmit ./test/type-test.ts && ts-node --compilerOptions '{\"module\":\"CommonJS\"}' ./test/type-test.ts",
```

feels more appropriate to me.

rollup.config.mjs
```diff
@@ -0,0 +1,68 @@
+import typescript from "@rollup/plugin-typescript";
+import terser from "@rollup/plugin-terser";
+
+const banner = `/*! loglevel - https://github.com/pimterry/loglevel - licensed MIT */`;
```
Ryuno-Ki, line 4: See my comment above on the shape of the banner.

Ryuno-Ki (commented):
Looks like it's on a good path. I feel like the typings could be more rigour.

src/loglevel.ts
wayfarer3130: Generally I think we should internally convert level to the number, while externally allowing both the name and the numbers to be passed.

